Issue #190, opened 2023-04-14
Get columns that get added/changed/removed by a `TableTransformer` 


lars-reimann:
### Is your feature request related to a problem?

It would be useful to know which columns get added, changed, or removed by a `TableTransformer`.

### Desired solution

The `TableTransformer` abstract base class should get three new **abstract** methods:

* `names_of_added_columns`
* `names_of_changed_columns`
* `names_of_removed_columns`

Each of them should return a list of strings. All subclasses of `TableTransformer` must implement them.

zzril:
It is unclear which table the methods should refer to.
E.g., should `get_names_of_changed_columns`, applied on an `Imputer` instance, return a list of those column names that have been changed in the last table the `Imputer` has been used to transform? (That would mean we store data that logically belongs to a `Table` globally in the `Transformer` class.)
Or should it return all the column names that can in principle be affected, i.e. all names that had been passed to `fit`?

lars-reimann:
> Or should it return all the column names that can in principle be affected, i.e. all names that had been passed to fit?

This.

PR #195, opened 2023-04-16
feat: `OneHotEncoder.inverse_transform` now maintains the column order from the original table

### LINKED ISSUE #109: `OneHotEncoder` should maintain column order

`OneHotEncoder.transform` currently appends the one-hot-encoded columns at the end of the result. Likewise, `OneHotEncoder.inverse_transform` appends the decoded columns at the end of the result. Because of this, `inverse_transform` is generally not able to restore the original table exactly.

**Example:**

_Input:_

|col1|col2|
|-|-|
|"a"|0|
|"b"|1|

_Result of transforming "a":_

|col2|col1_a|col1_b|
|-|-|-|
|0|1|0|
|1|0|1|

_Result of inverse transform:_

|col2|col1|
|-|-|
|0|"a"|
|1|"b"|

### Desired solution

Produce the following results instead:

_Result of transforming "a":_

|col1_a|col1_b|col2|
|-|-|-|
|1|0|0|
|0|1|1|

_Result of inverse transform:_

|col1|col2|
|-|-|
|"a"|0|
|"b"|1|

_No response_

### Screenshots (optional)

_No response_

_No response_

Marsmaennchen221:
Closes #109.

### Summary of Changes

`OneHotEncoder.inverse_transform` now maintains the column order from the original table (#109)
Fixed bug with `OneHotEncoder.inverse_transform` to not work if not all columns were fitted
New feature columns in `OneHotEncoder` will now be inserted where the combined columns were in the original table

### DIFF
--- a/src/safeds/data/tabular/transformation/_one_hot_encoder.py
+++ b/src/safeds/data/tabular/transformation/_one_hot_encoder.py
@@ -8,6 +8,7 @@ from safeds.data.tabular.exceptions import TransformerNotFittedError, UnknownCol
 from safeds.data.tabular.transformation._table_transformer import (
     InvertibleTableTransformer,
 )
+from safeds.data.tabular.typing import Schema
 
 
 class OneHotEncoder(InvertibleTableTransformer):
@@ -16,6 +17,7 @@ class OneHotEncoder(InvertibleTableTransformer):
     def __init__(self) -> None:
         self._wrapped_transformer: sk_OneHotEncoder | None = None
         self._column_names: list[str] | None = None
+        self._original_schema: Schema | None = None
 
     # noinspection PyProtectedMember
     def fit(self, table: Table, column_names: list[str] | None = None) -> OneHotEncoder:
@@ -50,6 +52,7 @@ class OneHotEncoder(InvertibleTableTransformer):
         result = OneHotEncoder()
         result._wrapped_transformer = wrapped_transformer
         result._column_names = column_names
+        result._original_schema = table.schema
 
         return result
 
@@ -120,12 +123,16 @@ class OneHotEncoder(InvertibleTableTransformer):
         data.columns = transformed_table.get_column_names()
 
         decoded = pd.DataFrame(
-            self._wrapped_transformer.inverse_transform(transformed_table._data),
+            self._wrapped_transformer.inverse_transform(transformed_table.keep_only_columns(self._wrapped_transformer.get_feature_names_out())._data),
             columns=self._column_names,
         )
         unchanged = data.drop(self._wrapped_transformer.get_feature_names_out(), axis=1)
 
-        return Table(pd.concat([unchanged, decoded], axis=1))
+        res = Table(pd.concat([unchanged, decoded], axis=1))
+        column_names = self._original_schema.get_column_names()
+        res = res.sort_columns(lambda col1, col2: column_names.index(col1.name) - column_names.index(col2.name))
+
+        return res
 
     def is_fitted(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


lars-reimann:
## [🦙 MegaLinter](https://megalinter.io/6.22.1) status: ❌ [ERROR](https://github.com/Safe-DS/Stdlib/actions/runs/4714957736)

| Descriptor  |                                 Linter                                 |Files|Fixed|                            Errors                            |Elapsed time|
|-------------|------------------------------------------------------------------------|-----|----:|--------------------------------------------------------------|------------|
|✅ PYTHON    |[black](https://megalinter.io/6.22.1/descriptors/python_black)          |    2|    2|                                                             0|0.76s       |
|❌ PYTHON    |[mypy](https://megalinter.io/6.22.1/descriptors/python_mypy)            |    2|     |[1](https://github.com/Safe-DS/Stdlib/actions/runs/4714957736)|1.66s       |
|❌ PYTHON    |[ruff](https://megalinter.io/6.22.1/descriptors/python_ruff)            |    2|    2|[2](https://github.com/Safe-DS/Stdlib/actions/runs/4714957736)|0.04s       |
|✅ REPOSITORY|[git_diff](https://megalinter.io/6.22.1/descriptors/repository_git_diff)|yes  |     |no                                                            |0.02s       |

See detailed report in [MegaLinter reports](https://github.com/Safe-DS/Stdlib/actions/runs/4714957736)

lars-reimann (changes requested):
Solving this issue is a little more complicated than that: Say you call `fit` with this table and run the encoder only on column `col1`:

|col1|col2|
|-|-|
|"a"|0|
|"b"|1|

You store the schema of this table in `self._original_schema`. Now you call `transform` with the following table:

|col2|col1|
|-|-|
|0|"a"|
|1|"b"|

If you now call `inverse_transform` on the result, you will get 

|col1|col2|
|-|-|
|"a"|0|
|"b"|1|

because the relative order of columns is now determined by the table the transformer was **fitted** with. This is not the table before transform, though.

A more thorough solution would be that `transform` already maintains the order of columns by replacing a column with its one-hot-encoded version instead of appending all new columns at the end. For example, calling `transform` on

|col2|col1|
|-|-|
|0|"a"|
|1|"b"|

should result in

|col2|col1_a|col1_b|
|-|-|-|
|0|1|0|
|1|0|1|

and calling `transform` on

|col1|col2|
|-|-|
|"a"|0|
|"b"|1|

should result in

|col1_a|col1_b|col2|
|-|-|-|
|1|0|0|
|0|1|1|

Then the `inverse_transform` can simply maintain the column order.

lars-reimann (changes requested):
I've added another test that should work but doesn't.

Marsmaennchen221:
> I've added another test that should work but doesn't.
> 
> Since the column name and the value could contain underscores, it's not possible to figure out which original column a column in the transformed table corresponds to. For example, `a_b_c` could correspond to the column called `a_b` (and value `c`) or a column called `a` (and value `b_c`).
> 
> Because of this, we'll need to store the mapping from original columns to transformed columns during `fit` (`dict[str, list[str]]`, with original column names as keys and names of corresponding transformed columns as values). This will also be useful for #190 later. During `transform` and `inverse_transform` you can then use this mapping to figure out how the columns need to be sorted instead of relying on the `startswith` check.

This would not solve the problem. If you have both columns in the original table the transformed table will contain only the column `a_b_c` once. One of the two columns would be ignored. Therefore i ignored this problem. You would have to make the column name distinctive from a possible transformed column name but I do not know a way to do this right now. Or you could raise a specified error to inform the user to change the column names.

lars-reimann:
I've created another issue (#201) about conflicting column names when using a `OneHotEncoder`. That can be done in a separate PR.

PR #303, opened 2023-05-12
feat: Add custom `OutOfBoundsError` and replace existing `ValueErrors`

### LINKED ISSUE #262: Custom error if number is out of bounds

We have several cases where we check that a number is in some interval. However, error messages vastly differ.

### Desired solution

* New error class `OutOfBoundsError` that inherits from `ValueError`
* The bounds and the actual value should be passed to the constructor as arguments
* The error message should then be build in the constructor and passed to the `super` constructor

_No response_

### Screenshots (optional)

_No response_

_No response_

alex-senger:
Closes #262.

### Summary of Changes

Add custom `OutOfBoundsError` and replace existing `ValueErrors` that need to be more specific.

### DIFF
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -1064,7 +1064,7 @@ class Table:
         Raises
         ------
         OutOfBoundsError
-            If the given percentage is smaller than 0 or larger than 1.
+            If the given percentage is smaller than or equal to 0 or larger than or equal to 1.
         """
         if percentage_in_first <= 0 or percentage_in_first >= 1:
             raise OutOfBoundsError(percentage_in_first, 0, 1)
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -26,6 +26,7 @@ from safeds.exceptions import (
     SchemaMismatchError,
     UnknownColumnNameError,
 )
+from safeds.exceptions._data import OutOfBoundsError
 
 from ._column import Column
 from ._row import Row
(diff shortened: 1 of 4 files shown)


src/safeds/exceptions/_data.py
```diff
@@ -92,3 +92,10 @@ class ValueNotPresentWhenFittedError(Exception):
 
     def __init__(self, value: str, column: str) -> None:
         super().__init__(f"Value not present in the table the transformer was fitted on: \n{value} in column {column}.")
+
+
+class OutOfBoundsError(ValueError):
+    """Exception raised when a value is not in the range [lower, upper]."""
+
+    def __init__(self, value: float, lower: float, upper: float) -> None:
+        super().__init__(f"Value {value} is not in the range [{lower}, {upper}].")
```
PhilipGutberlet, line 101: adjust the message for the boundaries of _table line 978-981.

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1053,10 +1061,13 @@ def split(self, percentage_in_first: float) -> tuple[Table, Table]:
             A tuple containing the two resulting tables. The first table has the specified size, the second table
             contains the rest of the data.
 
-
+        Raises
+        ------
+        OutOfBoundsError
+            If the given percentage is smaller than 0 or larger than 1.
```
PhilipGutberlet, line 1067: It also raises when the given percentage is equal to 0

alex-senger, line 1067:
```suggestion
            If the given percentage is smaller than or equal to 0 or larger than or equal to 1.
```

PR #308, opened 2023-05-19
feat: improve error handling of table

robmeth:
Closes #147.

### Summary of Changes

feat: Added `WrongFileExtensionError`

* Validate inputs of a function as the first step
* Raise an appropriate exception with an appropriate message if issues are found
* Document in the docstring that the exception is raised and under which condition
* Test that the correct exception with the correct message is raised

Co-authored-by: Marsmaennchen221 <47296670+Marsmaennchen221@users.noreply.github.com>

Issue #123, opened 2023-03-30
Check that methods of `Table` can handle an empty table


lars-reimann:
The methods of `Table` should work properly on an empty table. We don't currently handle this edge case, however. An appropriate tests should be added for each method of the `Table` class.

Example: It should be possible to add any row and any column to an empty table.

zzril:
What about the `Transformers`? Should it be possible to `fit` them on empty tables? Also, should it be possible to `transform` empty tables using them?  
(The latter one is probably always trivially possible.)

PR #309, opened 2023-05-19
test: Added test for `Table.remove_rows_with_outliers` to test multiple outliers in one column and outliers in two different columns

### LINKED ISSUE #274: Pytest can't detect 2 or more outliers in a function
### Describe the bug

The following function: _**def remove_rows_with_outliers**_ doesn't perform its function correctly because **it can only detect and remove a maximum of 1 outlier**. That is, 2 or more outliers should be removed but that doesn't work.

### To Reproduce

1.  Run the following function: **_test_should_remove_rows_with_outliers_** in _/Stdlib/tests/safeds/data/tabular/containers/table/test_remove_rows_with_outliers.py_ 

> The Tests should run successfully and remove only 1 row containing an outlier

2.  Add 1 or more outliers to an existing column e.g _col1, col2 or col3._ 

> Outlier Definition:  An outlier is defined as a value that has a distance of more than 3 standard deviations from the column mean.  Missing values are not considered outliers. They are also ignored during the calculation of the standard deviation.

3. Repeat step 1

### Expected behavior

_All Outliers should be detected and removed._

### Screenshots (optional)

_No response_

_No response_

Marsmaennchen221:
### Summary of Changes

test: Added test for `Table.remove_rows_with_outliers` to test multiple outliers in one column and outliers in two different columns

See my [Comment](https://github.com/Safe-DS/Stdlib/issues/274#issuecomment-1554492317) in the referenced issue for further explanation as there is no bug with the current method `Table.remove_rows_with_outliers`

### Additional Context

See #274


Marsmaennchen221:
Waiting for either #308 or #123 for the improved `Table.__eq__` method for the comparison of the empty tables.

alex-senger (approved):
LGTM

tests/safeds/data/tabular/containers/_table/test_remove_rows_with_outliers.py
```diff
@@ -78,7 +250,14 @@
...
     ],
-    ids=["no outliers", "with outliers", "no rows", "empty"],
+    ids=[
+        "no outliers",
+        "one outlier",
+        "outliers in two different columns",
+        "multiple outliers in one column",
+        "no rows",
+        "empty",
+    ],
```
jxnior01, line 260: LGTM. All edge cases covered.

jxnior01 (approved):

PR #311, opened 2023-05-19
feat:  Schema knows which columns have missing values

### LINKED ISSUE #44: Inferred schema should know which columns have missing values
### Is your feature request related to a problem? Please describe

The inferred schema of a `Table` currently treat columns of type `X` with and without missing values the same. For example, in the following table both columns would get type `int`:

|a|b|
|-|-|
|1|1|
|2| |

### Desired solution

1. There is already a class `OptionalColumnType` but it's not used. It should be used to clearly mark columns with missing values. 
2. Add an abstract method `is_missing` or `may_be_missing` to `ColumnType`.
3. Implement it for the subclasses.
4. Use this method to reimplement `Table.list_columns_with_missing_values` and `Column.has_missing_values`.

PhilipGutberlet:
Closes #44.

### Summary of Changes
added abstract method may_be_missing+
implemented the method in the subclasses+
reimplemented method has_missing_values in Column+ 
implemented list_columns_with_missing_values in Table

Co-authored-by: daniaHu <129186516+daniaHu@users.noreply.github.com>

### DIFF
--- a/tests/safeds/data/tabular/typing/test_column_type.py
+++ b/tests/safeds/data/tabular/typing/test_column_type.py
@@ -1,15 +1,6 @@
-
 import numpy as np
 import pytest
-from safeds.data.tabular.typing._column_type import (
-    Anything,
-    Boolean,
-    ColumnType,
-    Integer,
-    RealNumber,
-    String,
-    Optional
-)
+from safeds.data.tabular.typing._column_type import Anything, Boolean, ColumnType, Integer, Optional, RealNumber, String
 
 
 class TestFromNumpyDataType:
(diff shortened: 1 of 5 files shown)


tests/safeds/data/tabular/typing/test_column_type.py
```diff
@@ -106,8 +103,33 @@ class TestIsNumeric:
...
+            (Optional(is_nullable=False), False),
+            (Optional(is_nullable=True), False),
         ],
         ids=repr,
     )
     def test_should_return_whether_the_column_type_is_numeric(self, column_type: ColumnType, expected: bool) -> None:
         assert column_type.is_numeric() == expected
+
+
+class TestIsOptional:
```
lars-reimann, line 115:
```suggestion
class TestMayBeMissing:
```

lars-reimann (changes requested):
Not a complete review yet, just one observation.

PR #332, opened 2023-05-29
feat: Maintain tagging in methods inherited from `Table` class

zzril:
Closes #58.

### Summary of Changes

* Overrode methods in `TaggedTable` inherited from `Table` to maintain tagging.
* Introduced new exceptions to be thrown when the overriden method cannot be applied
* Added tests.

PR #345, opened 2023-06-04
feat: don't override when method could throw

zzril:
(Note: This PR is not to be merged into main, but into a feature branch.)

In `TaggedTable`, don't override those method of `Table` that could potentially remove the target column.  
Instead, introduce new methods that specifically promise to maintain tagging, and throw an exception otherwise.

See also [this comment](https://github.com/Safe-DS/Stdlib/pull/332#issuecomment-1575588526) in the original PR.


zzril:
Will not implement this, instead change the parent method's signature/docstring to throw the exception. See [this comment].

PR #385, opened 2023-06-23
feat: suggest similar columns if column gets accessed that doesnt exist

### LINKED ISSUE #203: Suggest similar columns if column gets accessed that doesn't exist

Sometimes people just mistype a column name.

### Desired solution

* We should see whether a column with a similar name exists in the current `Table` and include this in the error message ("did you mean XY"). To compute similarity of column names, use [`levenshtein`](https://pypi.org/project/Levenshtein/).

_No response_

### Screenshots (optional)

_No response_

If a column with the given name has existed in the past (#328), we should show that instead of looking for columns with similar names.

robmeth:
Closes #203

### Summary of Changes

* Added a method `_get_similar_columns` to find columns with a name similar to a given name.
* Updated `UnknownColumnNameError` to allow for passing name suggestions.
* Used these suggestions in `get_column`, `keep_only_columns`, `remove_columns`, `rename_column`, `replace_column`, `transform_column`, `plot_lineplot`, `plot_scatterplot`.

Co-authored-by: jxnior01 <129027012+jxnior01@users.noreply.github.com>


lars-reimann:
## [🦙 MegaLinter](https://megalinter.io/7.0.4) status: ❌ [ERROR](https://github.com/Safe-DS/Stdlib/actions/runs/5356385906)

| Descriptor  |                                Linter                                 |Files|Fixed|                            Errors                            |Elapsed time|
|-------------|-----------------------------------------------------------------------|-----|----:|--------------------------------------------------------------|------------|
|✅ PYTHON    |[black](https://megalinter.io/7.0.4/descriptors/python_black)          |    2|    2|                                                             0|1.81s       |
|❌ PYTHON    |[mypy](https://megalinter.io/7.0.4/descriptors/python_mypy)            |    2|     |[1](https://github.com/Safe-DS/Stdlib/actions/runs/5356385906)|2.26s       |
|❌ PYTHON    |[ruff](https://megalinter.io/7.0.4/descriptors/python_ruff)            |    2|    2|[1](https://github.com/Safe-DS/Stdlib/actions/runs/5356385906)|0.03s       |
|✅ REPOSITORY|[git_diff](https://megalinter.io/7.0.4/descriptors/repository_git_diff)|yes  |     |no                                                            |0.01s       |

See detailed report in [MegaLinter reports](https://github.com/Safe-DS/Stdlib/actions/runs/5356385906)

tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py
```diff
@@ -0,0 +1,18 @@
...
+    [
+        (Table({"column1": ["col1_1"], "col2": ["col2_1"], "cilumn2": ["cil2_1"]}), "col1", ["col2"]),
+        (Table({"column1": ["col1_1"], "col2": ["col2_1"], "cilumn2": ["cil2_1"]}), "clumn1", ["column1", "column3"]),
+    ],
+    ids=["one similar", "two similar"],
+)
+def test_should_warn_if_similar_column_name(table1: Table, column_name: str, expected: list[str]) -> None:
+    with pytest.warns(
+        UserWarning,
+        match=f"The Column {column_name} is unknown.\n Did you mean one of these: {expected}?",
```
Marsmaennchen221, line 16: The message is not being tested correctly. The square brackets from the list act as regex characters. You can instead use something like this:
```suggestion
        match=rf"The Column {column_name} is unknown.\n Did you mean one of these: \[{str(expected)[1:-1]}\]?",
```
You also have to correct the tests as the expected column names are not the ones your method produces

Marsmaennchen221 (changes requested):

src/safeds/data/tabular/containers/_table.py
```diff
@@ -523,6 +525,35 @@ def get_row(self, index: int) -> Row:
 
         return Row._from_pandas_dataframe(self._data.iloc[[index]], self._schema)
 
+    def get_similar_columns(self, column_name: str) -> bool:
```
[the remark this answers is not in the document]
lars-reimann: The method should also not be public. It's just used internally for better error messages.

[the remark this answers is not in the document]
lars-reimann: Likewise, I would not expect that a method call `get_similar_columns` returns a `bool`, as pointed out by @Marsmaennchen221. Neither would I expect that it warns.

src/safeds/data/tabular/containers/_table.py
```diff
@@ -691,6 +693,27 @@ def get_row(self, index: int) -> Row:
 
         return Row._from_pandas_dataframe(self._data.iloc[[index]], self._schema)
 
+    def _get_similar_columns(self, column_name: str) -> list[str]:
+        """
+        Give a warning with similar columns to the given one that's not in the table.
```
Marsmaennchen221, line 698:
```suggestion
        Get all similar column names to the given one that's not in the table.
```

src/safeds/data/tabular/containers/_table.py
```diff
@@ -691,6 +693,27 @@ def get_row(self, index: int) -> Row:
...
+
+        Parameters
+        ----------
+        column_name : str
+            The name of the Column, that's not in the Table.
+
+        Returns
+        -------
+        bool
+            True if there are similar columns.
```
Marsmaennchen221, line 708:
```suggestion
        similar_columns: list[str]
            A list of all similar column names.
```

src/safeds/exceptions/_data.py
```diff
@@ -16,8 +16,8 @@ class UnknownColumnNameError(KeyError):
         The name of the column that was tried to be accessed.
     """
 
-    def __init__(self, column_names: list[str]):
-        super().__init__(f"Could not find column(s) '{', '.join(column_names)}'")
+    def __init__(self, column_names: list[str], similar_columns: list[str] | None = None):
+        super().__init__(f"Could not find column(s) '{', '.join(column_names)}'. \nDo you mean '{similar_columns}")
```
Marsmaennchen221, line 20:
```suggestion
        super().__init__(f"Could not find column(s) '{', '.join(column_names)}'. \nDid you mean '{similar_columns}'?")
```

src/safeds/exceptions/_data.py
Marsmaennchen221, line 1: Please test the behaviour of the exception with the similar columns

Marsmaennchen221 (changes requested):

Marsmaennchen221 (approved):

src/safeds/data/tabular/containers/_table.py
```diff
@@ -695,6 +697,27 @@ def get_row(self, index: int) -> Row:
...
+            The name of the Column, that's not in the Table.
+
+        Returns
+        -------
+        similar_columns: list[str]
+            A list of all similar column names.
+        """
+        similar_columns = []
+        for column in self.column_names:
+            if Levenshtein.jaro_winkler(column, column_name) >= 0.7:
```
zzril, line 716: Is there a specific reason for this exact threshold? (Like, did you test it and it gave the best results, or is this a generally a "common" value used in such a case?)

Could also set some lower and upper limit of columns to return. (Like, always return at least one, but never more than 3.)

zzril:
Should probably also document in the PR description what methods make use of the new functionality.

src/safeds/data/tabular/containers/_table.py
robmeth: the 0.7 was just a random number because i couldn't find any "commun" values.

I don't think there should be a limit to the returned columns because there could be no columns that are similar enough or of there are many similar columns but the one needed isn't in there. I think it's more useful to have all of them there.
I'll try dynamically increasing the threshold

tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+from safeds.data.tabular.containers import Table
+from safeds.exceptions._data import UnknownColumnNameError
+
+
+@pytest.mark.parametrize(
+    ("table", "column_name", "expected"),
+    [
+        (Table({"column1": ["col1_1"], "col2": ["col2_1"], "cilumn2": ["cil2_1"]}), "col1", ["column1", "col2"]),
+        (Table({"column1": ["col1_1"], "col2": ["col2_1"], "cilumn2": ["cil2_1"]}), "clumn1", ["column1", "cilumn2"]),
```
zzril, line 10: May also want to include testcases where some of the column names are way off, so it's easier to see what columns are the "right" ones to suggest.

zzril (changes requested):

tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py
robmeth: done

src/safeds/data/tabular/containers/_table.py
zzril: My main concern are tables with a lot of similar names, for example those created by one-hot-encoding a column with a lot of different values.

Getting an error messgage like `Did you mean age__8, age__9, age__10, age__11, ..., age_81` is probably not the most helpful.

Maybe sthg like:
* Create a dictionary/mapping of column names --> similarity
* sort by similarity
* if 1st one has similarity >= 0.5: take into suggestions; else break
* if 2nd one has similarity >= 0.7: take into suggestions; else break
* if 3rd one has similarity >= 0.9: take into suggestions
* return suggestions

We can do fine-tuning later, but this should be good enough to prevent the pathological cases.

src/safeds/data/tabular/containers/_table.py
```diff
@@ -695,6 +697,34 @@ def get_row(self, index: int) -> Row:
 
         return Row._from_pandas_dataframe(self._data.iloc[[index]], self._schema)
 
+    def _get_similar_columns(self, column_name: str) -> list[str]:
+        """
+        Get all similar column names to the given one that's not in the table.
```
zzril, line 702: Since we also return names that match the given name exactly:

```suggestion
        Get all the column names in a Table that are similar to a given name.
```

src/safeds/data/tabular/containers/_table.py
```diff
@@ -695,6 +697,34 @@ def get_row(self, index: int) -> Row:
...
+
+        Parameters
+        ----------
+        column_name : str
+            The name of the Column, that's not in the Table.
```
zzril, line 707: (Same reason as above.)
```suggestion
            The name to compare the Table's column names to.
```

src/safeds/data/tabular/containers/_table.py
```diff
@@ -695,6 +697,34 @@ def get_row(self, index: int) -> Row:
...
+
+        Returns
+        -------
+        similar_columns: list[str]
+            A list of all similar column names.
```
zzril, line 712: Maybe be a bit more exact here:
```suggestion
            A list of all column names in the Table that are similar or equal to the given column name.
```

zzril (dismissed):

src/safeds/exceptions/_data.py
```diff
@@ -16,8 +16,18 @@ class UnknownColumnNameError(KeyError):
...
+    def __init__(self, column_names: list[str], similar_columns: list[str] | None = None):
+        class _UnknownColumnNameErrorMessage(
+            str,
+        ):  # This class is necessary for the newline character in a KeyError exception. See https://stackoverflow.com/a/70114007
+            def __repr__(self) -> str:
+                return str(self)
+
+        super().__init__(
+            _UnknownColumnNameErrorMessage(
+                f"Could not find column(s) '{', '.join(column_names)}'.\nDid you mean '{similar_columns}'?",
```
zzril, line 28: If `_get_similar_columns()` did not find any similar columns, we should probably omit the `Did you mean`.

Could also leave the `similar_columns` as an optional parameter and print the old message if that parameter `None`.

tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py
```diff
@@ -0,0 +1,35 @@
...
+                    "col4": ["col2_1"],
+                    "cilumn2": ["cil2_1"],
+                },
+            ),
+            "clumn1",
+            ["column1", "cilumn2"],
+        ),
+    ],
+    ids=["one similar", "two similar/ dynamic increase"],
+)
```
zzril, line 25: We also need a case for an empty table.

And I guess it also makes sense to have a test case that *returns* an empty list.

zzril:
Updated some typos in the PR description.

Aside from the things mentioned in the review, this looks fine.

src/safeds/exceptions/_data.py
jxnior01: Thanks. We'll update the code with an `if similar_columns is not None: error_message += f"\nDid you mean '{similar_columns}'?"`

tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py
jxnior01: Thanks. It will be updated.

lars-reimann:
Regarding the conflicts in `pyproject.toml` and `poetry.lock`:

1. Accept the changes from `main`
2. Run `poetry add levenshtein` again

zzril:
One - hopefully last - remark:

We're using the `_get_similar_columns` method only before calls to `UnknownColumnNameError`, in order to construct the correct list of suggestions.

Would it make sense to change the design like this?
* dop the `similar_columns: list[str]` parameter from `UnknownColumnNameError.__init__(...)`
* add a (keyword) parameter: `suggest_columns: bool` that is set to `True` by default
* if `suggest_columns` is set to true, build the suggestions list *inside* the `UnknownColumnNameError`
* remove the (now unnecessary) construction of this list in the corresponding table functions

zzril:
> @zzril to call the method i'd have to import Table in `_data.py` and somehow that doesn't work
> 
> > (most likely due to a circular import)

You need to do the import inside the `TYPE_CHECKING` block:
```py
if TYPE_CHECKING:
    from pathlib import Path
    from safeds.data.tabular.containers import Table
```

For the method itself, maybe sthg like this:

```py
class UnknownColumnNameError(KeyError):
    """
    Exception raised for trying to access an invalid column name.

    Parameters
    ----------
    column_names : list[str]
        The name of the column that was tried to be accessed.
    """

    def __init__(self, column_names: list[str], *, suggest_similar_names_from: Table = None):
        class _UnknownColumnNameErrorMessage(
            str,
        ):  # This class is necessary for the newline character in a KeyError exception.
            # See https://stackoverflow.com/a/70114007
            def __repr__(self) -> str:
                return str(self)

        error_message = f"Could not find column(s) '{', '.join(column_names)}'."

        if suggest_similar_names_from is not None:
            similar_columns = []
            for column_name in column_names:
                similar_columns += suggest_similar_names_from._get_similar_columns(column_name)
            if len(similar_columns) > 0:
                error_message += f"\nDid you mean '{similar_columns}'?"

        super().__init__(_UnknownColumnNameErrorMessage(error_message))
```

This way, other code can just ignore the feature and still work fine, or explicitely ask for the suggestions to be printed.

(Would need to adapt a few tests then.)

src/safeds/exceptions/_data.py
zzril: Better check that
1. `similar_columns is not None`
2. `len(similar_columns` > 0
Only then add it to the error message.

src/safeds/exceptions/_data.py
```diff
@@ -16,8 +16,19 @@ class UnknownColumnNameError(KeyError):
...
+    def __init__(self, column_names: list[str], similar_columns: list[str] | None = None):
+        class _UnknownColumnNameErrorMessage(
+            str,
+        ):  # This class is necessary for the newline character in a KeyError exception. See https://stackoverflow.com/a/70114007
+            def __repr__(self) -> str:
+                return str(self)
+
+        error_message = f"Could not find column(s) '{', '.join(column_names)}'."
+
+        if similar_columns is not None:
```
zzril, line 28:
```suggestion
        if similar_columns is not None and len(similar_columns) > 0:
```

tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py
```diff
@@ -0,0 +1,46 @@
...
+                },
+            ),
+            "clumn1",
+            ["column1", "cilumn2"],
+        ),
+        (
+            Table({"column1": ["col1_1"], "x": ["y"], "cilumn2": ["cil2_1"]}),
+            "notexisting",
+            ["x"],
+        ),
```
zzril, line 27: This doesn't match the ID. The ID says `no similar`.

```suggestion
               (
            Table({"column1": ["a"], "column2": ["b"], "column3": ["c"]}),
            "notexisting",
            [],
        ),
```

src/safeds/exceptions/_data.py
jxnior01: Ohh I thought this was resolved. Looking into it.

jxnior01: We already have such a test case. Have you seen the second test case in `tests/safeds/data/tabular/containers/_table/test_get_similar_columns.py`? Except we'll have to move it, which i find not necessary.

Marsmaennchen221 (dismissed):

src/safeds/exceptions/_data.py
zzril: That's a test for `_get_similar_columns` directly.
I was talking abour a test case that ensures that the `UnknownColumnNameError` can be passed an empty list in the constructor.

zzril:
Converning the update of `poetry.lock`: Is this safe?

When I updated this earlier, I only (automatically) added the dependencies for `levenshtein`.

But now I see a later commit has also re-generated it with a newer poetry version and added the `category` in every entry. Is this necessary? @lars-reimann @Marsmaennchen221

tests/safeds/exceptions/test_unknown_column_name_error.py
```diff
@@ -0,0 +1,15 @@
...
+@pytest.mark.parametrize(
+    ("column_names", "similar_columns", "expected_error_message"),
+    [
+        (["column1", "column2"], [], "Could not find column(s) 'column1, column2'."),
+    ],
+    ids=["empty_list"]
+)
+def test_empty_similar_columns(column_names, similar_columns, expected_error_message):
+    error = UnknownColumnNameError(column_names, similar_columns)
+    assert str(error) == expected_error_message
```
zzril, line 14: We generally use `with pytest.raises` for that.

Will make a suggestion in a moment.

src/safeds/exceptions/_data.py
jxnior01: Please review the added file.

tests/safeds/exceptions/test_unknown_column_name_error.py
zzril, line 14:
```suggestion
@pytest.mark.parametrize(
    ("column_names", "similar_columns", "expected_error_message"),
    [
        (["column1", "column2"], [], r"Could not find column\(s\) 'column1, column2'."),
    ],
    ids=["multiple_columns_no_suggestions"]
)
def test_empty_similar_columns(column_names: list[str], similar_columns: list[str], expected_error_message: str):
    with pytest.raises(UnknownColumnNameError, match=expected_error_message):
        raise UnknownColumnNameError(column_names, similar_columns)
```

jxnior01: I just made some changes and resolved linter errors. I guess it's similar to yours. See last commit =)

zzril:
> > Concerning the update of `poetry.lock`: Is this safe?
> > When I updated this earlier in [6419940](https://github.com/Safe-DS/Stdlib/commit/641994089e34b1b954753218344c7148789fce81), I only (automatically) added the dependencies for `levenshtein`.
> > But now I see a later commit has also re-generated it with a newer poetry version and changed the `category` in every entry. Is this necessary? @lars-reimann @Marsmaennchen221
> 
> I'd suggest reverting this file back to the state of [6419940](https://github.com/Safe-DS/Stdlib/commit/641994089e34b1b954753218344c7148789fce81). When a new dependency gets added, the lockfile should only be changed minimally.

Done.

tests/safeds/exceptions/test_unknown_column_name_error.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from safeds.exceptions import UnknownColumnNameError
+
+
+@pytest.mark.parametrize(
+    ("column_names", "similar_columns", "expected_error_message"),
+    [
+        (["column1", "column2"], [], r"Could not find column\(s\) 'column1, column2'."),
+    ],
+    ids=["empty_list"],
```
zzril, line 10: Let me add some more test cases here...

Marsmaennchen221 (approved):

lars-reimann (approved):
The changes to `pyproject.toml` and `poetry.lock` are OK. I didn't review the rest.

PR #398, opened 2023-06-27
feat: Rangescalar and other transformer implementations now pass exceptions from sklearn

### LINKED ISSUE #326: `RangeScaler` should not pass exception from scikit-learn when `fit`ting on empty table

See title.

### Desired solution

* Should throw new, user-friendly exception.
* Verify that exception is thrown.

Should also check if any the other `Transformer` implementations pass on exceptions from scikit-learn.

jxnior01:
Closes #326 

### Summary of Changes

* Rangescalar and other transformer implementations now pass exceptions from sk-learn when fitting on an empty table

### DIFF
--- a/src/safeds/data/tabular/transformation/_imputer.py
+++ b/src/safeds/data/tabular/transformation/_imputer.py
@@ -4,6 +4,7 @@ import warnings
 from typing import Any
 
 import pandas as pd
+import sklearn.exceptions as sk_exceptions
 from sklearn.impute import SimpleImputer as sk_SimpleImputer
 
 from safeds.data.tabular.containers import Table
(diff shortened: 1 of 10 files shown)


zzril:
We *don't* want the exceptions from scikit-learn to be passed on to the users. Instead, we want understandable, user-friendly exceptions.

The original issue was about one case where such an exception from sklearn *was* (unintentionally) passed through.

I'll need to check if that one has already been fixed by now. If so, I'll close the issue.

zzril:
Yeah, I guess the title was somewhat misleading. I thought I had given it a "bug" label, in which case the old title would have made for sense.
Anyway, changed the title now. Sorry for the confusion.

PR #408, opened 2023-06-30
test: unified instances of empty table in all tests

### LINKED ISSUE #347: style: Inconsistency in creating empty tables in tests

There's some inconsistency as to how create empty tables in test cases. Most of the time, we're using `Table()`, sometimes we use `Table({})`.

The latter is used in:
* `test_add_column` (2 times)
* `test_add_columns` (2x)
* `test_from_rows.py` (first set of parametrize-testcases)
* `test_remove_columns_with_missing_values` (2x)
* `test_remove_columns_with_non_numerical_values` (2x)
* `test_remove_rows_with_missing_values` (2x)
* `test_repr_html` (3x)
* `test_to_dict`
* `test_to_html`

Do we care about this?

If so, should change these cases to also use `Table()`, so it's the same everywhere.

### Desired solution

Change the above cases to use `Table()`.

Decide that we don't care.

daniaHu:
Closes #347

### Summary of Changes

Searched through tests for other forms of creating empty table than our agreed on standard "Table()", and switched them to the standard.


jxnior01 (changes requested):
Looks good. Just a single test: `test_remove_columns_with_missing_values` needs the change.

Marsmaennchen221 (changes requested):
Looks good so far.

You mised a few files:
_table/test_init.py
_table/test_number_of_columns.py
_table/test_number_of_rows.py
_table/test_remove_columns_with_missing_values.py

patrikguempel (approved):

Marsmaennchen221 (dismissed):
[LGTM](https://lgtm.party/p/RR17QLA/)

tests/safeds/data/tabular/containers/_table/test_init.py
```diff
@@ -8,7 +8,7 @@
     ("table", "expected"),
     [
         (Table(), Schema({})),
-        (Table({}), Schema({})),
+        (Table(), Schema({})),
```
zzril, line 11: The IDs suggest that this test is indeed supposed to distinguish between both cases. (Or rather, verify that both create the same table.)

zzril, line 11: The IDs suggest that this test is indeed supposed to distinguish between both cases. (Or rather, verify that both create the same table.)

zzril:
May also consider doing the same for `Row()` vs `Row({})`.

But can also leave that for another micro-issue. :D

tests/safeds/data/tabular/containers/_table/test_init.py
Marsmaennchen221, line 11:
```suggestion
        (Table({}), Schema({})),
```
@zzril Good point 👍

zzril (approved):

PR #419, opened 2023-07-05
docs: Image Processing tutorial

### LINKED ISSUE #373: Tutorial for image processing
There should be a notebook-based documentation page that describes the different image processing methods, including an example for each.

robmeth:
Closes #373

### Summary of Changes

There is a notebook-based documentation page that describes the different image processing methods, including an example for each.


zzril:
When generating the documentation from this branch with `poetry run mkdocs serve`, I cannot find the tutorials on the (localhost) webpage.
When I click on `Tutorials` on the top navigation bar, it still only lists `Data processing`, `Data visualization` and `Machine Learning`.

Looks like we need to add a reference to the new tutorial somewhere. Where?

jxnior01 (dismissed):
Looks good. Just a few grammatical errors:
Change: All operations on `a` Image return a new Image `to` All operations on `an` Image return a new Image.

docs/tutorials/image_processing.ipynb
jxnior01, line 1: In the `Blur the image` section: The radius defines the amount of pixels `combined` in each direction. I think `combined` is a better word to use.

zzril:
> Looks good. Just a few grammatical errors: Change: All operations on `a` Image return a new Image `to` All operations on `an` Image return a new Image.

Fixed.

docs/tutorials/image_processing.ipynb
zzril: I agree. changed it.

zzril:
Note to myself: Need to add `adjust_color_balance` to the tutorial.

zzril:
When generating docs with `poetry run mkdocs serve` and testing in a browser, the `Tutorials` page ages to load now. (Even before selecting the `Image Processing` tab.)

@lars-reimann Any ideas?

lars-reimann:
> When generating docs with `poetry run mkdocs serve` and testing in a browser, the `Tutorials` page takes ages to load now. (Even before selecting the `Image Processing` tab.)
> 
> @lars-reimann Any ideas?

Did the site fully compile already when you open it in the browser (see command line)?

zzril:
> > When generating docs with `poetry run mkdocs serve` and testing in a browser, the `Tutorials` page takes ages to load now. (Even before selecting the `Image Processing` tab.)
> > @lars-reimann Any ideas?
> 
> Did the site fully compile already when you open it in the browser (see command line)? For me the compilation takes 42 seconds but the site is responsive afterwards.

Looks like I didn't. Or maybe I accidentally changed something while I was loading the page, causing it to recompile.

Works fine when I wait until compilation is over.

docs/tutorials/image_processing.ipynb
Marsmaennchen221, line 1: We could use the following html code instead of the "Important Note" to display the message as a note admonition in mkdocs:
```html
<div class="admonition note">
  <p class="admonition-title">Note</p>
  <p>
    All operations on an `Image` return a new `Image`. The original `Image` will not be changed.
  </p>
</div>
```

lars-reimann: > Maybe @lars-reimann wants to have a say about this, though. I'm not sure if there's potentially any downsides in using hardcoded html inside our markdown.

The downside would be that we now depend on some internal details of `Material for Mkdocs`. Their CSS might change in the future.

> By the way, the html does not get rendered properly when viewing the file in GitHub (probably for security reasons). But I don't think that's much of a concern as there's rarely a need to open the files there.

That should not concern us and it's the same for admonitions in Markdown. The documentation is meant to be viewed on Read the docs.

zzril: Alright, I would suggest that we go with the html version then.

If problems start arising because `Material for Mkdocs` changes things too often, we can always go back to pure markdown.

Marsmaennchen221 (approved):
[LGTM](https://lgtm.party/p/NNRI9CG1/)

zzril:
As we already have some performance issues with the docs (#447), I'm wondering if adding a bunch of images will make things worse.

@lars-reimann Shall I merge now anyway, or do you prefer to first solve the performance issue?

lars-reimann:
That page loads quickly enough for me.

Issue #150, opened 2023-04-02
Improve error handling of `TaggedTable`


lars-reimann:
### Is your feature request related to a problem?

In some cases, error handling is not ideal (see solution for details).

### Desired solution

* Validate inputs of a function as the first step
* Raise an appropriate exception with an appropriate message if issues are found
* Document in the docstring that the exception is raised and under which condition
* Test that the correct exception with the correct message is raised

alex-senger:
Blocked by #58

guenterk:
I do not see why #58 blocks this issue. Only testing should be blocked until #58 is completed, to avoid useless double work. There is no reason why #58 prevents input validation, raising exceptions and documenting them. If all of that that is already implemented, mention it in the comment.

zzril:
Currently, `TaggedTable` has barely any methods yet.

There are the constructor-like methods `_from_table()` and `__init__()`, which do some sort of input validation. (Making sure that the `target` column won't be a `feature` column at the same time; not sure what else to validate here?)  
The only other methods are mere getters (`features` and `target`).

So, currently there's not much (possibly nothing?) to do in terms of "improving error handling" / input validation.

With #58 done, `TaggedTable` will have a whole lot of new methods, namely those ones inherited (and overridden) from `Table`.  
The solution implemented in the current PR for #58 (#332) does some input validation for these new methods already.

Once that PR is merged, once can think about refining the error handling. (Or alternatively make it part of #58 and drop this separate issue altogether.)

PR #420, opened 2023-07-05
docs: Standardize optional type in docstrings

### LINKED ISSUE #374: Standardize Optional Type in docstrings
Currently a parameter or return type of `some_type | None` is described in the docstring as either `some_type | None` or `Optional[some_type]`. We should set our standard for this case and change all other cases to this new standard.

Additionally there are cases where the parameter allows a None type but the docstring type does not show this. We should add this information to all those cases.

## Examples
`Optional[type]` in the `Imputer`
```py
 def fit(self, table: Table, column_names: list[str] | None) -> Imputer:
        """
        [...]
        Parameters
        ----------
        [...]
        column_names : Optional[list[str]]
            The list of columns from the table used to fit the transformer. If `None`, all columns are used.
        [...]
        """
```
`type | None` in `Table`:
```py
    def __init__(self, data: Mapping[str, Sequence[Any]] | None = None) -> None:
        """
        Create a table from a mapping of column names to their values.

        Parameters
        ----------
        data : Mapping[str, Sequence[Any]] | None
            The data. If None, an empty table is created.
        [...]
        """
```
Example of a method without this information in `Table.py`:
```py
def slice_rows(
        self,
        start: int | None = None,
        end: int | None = None,
        step: int = 1,
    ) -> Table:
        """
        Slice a part of the table into a new table.

        This table is not modified.

        Parameters
        ----------
        start : int
            The first index of the range to be copied into a new table, None by default.
        end : int
            The last index of the range to be copied into a new table, None by default.
        [...]
        """
```

jxnior01:
Closes #374 

### Summary of Changes

* Selected `optional type` and standardized it in docstrings


Marsmaennchen221:
Didn´t we say we want the docs to be the same as the implementation so we don´t want `Optional[...]` and we do want `... | None` in the implementation as well as in the docstrings?

docs/development/guidelines.md
```diff
@@ -265,6 +265,32 @@ The docstrings **should** use the [numpydoc](https://numpydoc.readthedocs.io/en/
...
+    ```
+
+!!! failure "**DON'T** (library code):"
+
+    ```py
+    def __init__(self, data: Mapping[str, Any] | None = None) -> None:
+        """
+        data : Optional[Mapping[str, Any]]
+        """
+
```
Marsmaennchen221, line 284:
```suggestion
    ```

```

docs/development/guidelines.md
```diff
@@ -265,6 +265,32 @@ The docstrings **should** use the [numpydoc](https://numpydoc.readthedocs.io/en/
...
+!!! failure "**DON'T** (library code):"
+
+    ```py
+    def __init__(self, data: Mapping[str, Any] | None = None) -> None:
+        """
+        data : Mapping[str, Any] | None
+        """
+
```
Marsmaennchen221, line 292:
```suggestion
    ```
```

Marsmaennchen221 (changes requested):

Marsmaennchen221 (changes requested):

alex-senger (changes requested):
There is still an "optional docstring" in `InvertibleTableTransformer.fit()` in `_table_transformer.py`

alex-senger (approved):
[LGTM](https://lgtm.party/p/1N7T2VRO/)

Marsmaennchen221 (changes requested):
Looks good so far.

You missed the docstrings in method `Table.slice_rows`

In `regression.AddaBoost` please change the docstring for `AdaBoost.learner` to return a `Regressor`

jxnior01:
> > Looks good so far.
> > You missed the docstrings in method `Table.slice_rows`
> > In `regression.AddaBoost` please change the docstring for `AdaBoost.learner` to return a `Regressor`
> 
> Adressed these issues. Also merged current `main` into this.

Thanks. I'm rounding up with issue #150. I planned resolving all reviews when i'm done.

Marsmaennchen221 (approved):
[LGTM](https://lgtm.party/p/2MISV3O/)

PR #425, opened 2023-07-06
fix: Warnings raised by underlying seaborn and numpy libraries 

### LINKED ISSUE #357: Warnings raised by underlying `seaborn` and `numpy` libraries

When running the test suite, currently 3 warnings will be raised from underlying libraries (`seaborn` and `numpy`) and are not catched.

These are the corresponding testcases:
* `tests/safeds/data/tabular/containers/_table/test_plot_correlation_heatmap.py::test_should_match_snapshot[empty]`
* `tests/safeds/data/tabular/containers/_table/test_plot_correlation_heatmap.py::test_should_match_snapshot[empty]`
* `tests/safeds/data/tabular/containers/_table/test_plot_histograms.py::test_should_match_snapshot[four columns]`

For the first 2, the warnings are raised by some `seaborn` module and seems to have the same cause, the 3rd one is some `numpy` thing.

### Desired solution

Ideal solution: Find the cause of the warnings and change the code to not produce these warnings anymore.

If it's absolutely clear that we can't do anything to fix these warnings:  
Silence them in the code that produced them. (Not just in the tests!)

jxnior01:
Closes #357 

### Summary of Changes

* Python's `warnings` module now ignores warnings caused by underlying seaborn and numpy libraries.


src/safeds/data/tabular/containers/_table.py
```diff
@@ -1774,16 +1774,31 @@ def plot_correlation_heatmap(self) -> Image:
...
-            data=only_numerical._data.corr(),
-            vmin=-1,
-            vmax=1,
-            xticklabels=only_numerical.column_names,
-            yticklabels=only_numerical.column_names,
-            cmap="vlag",
-        )
-        plt.tight_layout()
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
```
zzril, line 1778: You can also use a regex here (that matches the text with `xlim` and `ylim`), makes it easier to read.

zzril (changes requested):

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1774,16 +1774,31 @@ def plot_correlation_heatmap(self) -> Image:
...
+                "ignore",
+                message=(
+                    "Attempting to set identical low and high xlims makes transformation singular; automatically"
```
zzril, line 1781: Instead of using to separate filters, you can also use a regex here (that matches both the text with `xlim` and `ylim`). Makes it easier to read.

zzril:
Looks good to me, aside from the redundancy.

---

Generally, it would be good to understand what originally caused the warnings, though.

Maybe someone with a better knowledge of `numpy` and `matplotlib` should have a look at this and make sure we're not missing a better way to implement the parts that raised the warnings?
@lars-reimann @Marsmaennchen221

jxnior01:
> Looks good to me, aside from the redundancy.
> 
> Generally, it would be good to understand what originally caused the warnings, though.

I think the warning message is raised by the `sns.heatmap()` function from the seaborn library. It indicates that when it attempts to set identical low and high values for the x-axis (xlims) or y-axis (ylims) limits, it can result in a singular transformation. This is why seaborn automatically `expands the limits` to avoid this issue.

Marsmaennchen221:
As far as I understand the warning is only raised if you use an empty table. Maybe it would be better to catch the warning **only** if the user tries to use this function on an empty table and raise a new warning that informs the user that a correlation heatmap on an empty table will show nothing.

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1966,17 +1974,22 @@ def plot_histograms(self) -> Image:
...
+            grid = sns.FacetGrid(data=data, col="variable", col_wrap=col_wrap, sharex=False, sharey=False)
+            grid.map(sns.histplot, "value")
+            grid.set_xlabels("")
+            grid.set_ylabels("")
+            grid.set_titles("{col_name}")
+            for axes in grid.axes.flat:
+                axes.set_xticks(axes.get_xticks())
+                axes.set_xticklabels(axes.get_xticklabels(), rotation=45, horizontalalignment="right")
+            grid.tight_layout()
+            fig = grid.fig
```
Marsmaennchen221, line 1992:
```suggestion
        data = pd.melt(self._data.applymap(lambda value: str(value)), value_vars=self.column_names)
        grid = sns.FacetGrid(data=data, col="variable", col_wrap=col_wrap, sharex=False, sharey=False)
        grid.map(sns.histplot, "value")
        grid.set_xlabels("")
        grid.set_ylabels("")
        grid.set_titles("{col_name}")
        for axes in grid.axes.flat:
            axes.set_xticks(axes.get_xticks())
            axes.set_xticklabels(axes.get_xticklabels(), rotation=45, horizontalalignment="right")
        grid.tight_layout()
        fig = grid.fig
```
For `Table.plot_histograms` maybe we should replace all the data with strings. This would fix the warning and it also would make the plot look a lot better.
An example with the table `Table({"a": [1, 2, 3, 3, 2, 4, 3], "b": [True, True, False, True, False, False, True], "c": ["a", "b", "b", "b", "b", "b", "b"]})`:
Before:
original
And after:
all values as strings

Marsmaennchen221: This would also fix an error that is raised when you use a table with a column that contains boolean values and `None` values

patrikguempel (approved):

patrikguempel (changes requested):
nvm 
[!https://media.tenor.com/MnDVObkGxh8AAAAM/the-simpsons-nevermind.gif]

tests/safeds/data/tabular/containers/_table/test_plot_correlation_heatmap.py
Marsmaennchen221, line 1: Please test in a seperate method that the new warning is raised if you use an empty table

Marsmaennchen221 (changes requested):

Marsmaennchen221 (changes requested):

tests/safeds/data/tabular/containers/_table/test_plot_correlation_heatmap.py
Marsmaennchen221: Look at my suggestion below. You currently filter the warning (in all cases) and dont test if the warning is raised whith an empty table. You can accept my suggestion below as this covers all I meant with this comment :)

PR #430, opened 2023-07-07
feat: 381 blend & gaussian noise

patrikguempel:
Closes #381 

### Summary of Changes

Images can now be modified with "gaussian noise". Added blend method in the process.

Issue #12, opened 2023-01-20
Global seed variable


Gerhardsa0:
### Is your feature request related to a problem? Please describe

For functions like ```Table.shuffle()``` who are not deterministic we do not have the ability to set a seed and get always the same but random output.

### Desired solution

We should implement a global variable which sets on all non-deterministic functions a seed, so we get always the same output for those functions.

Also all tests/functions should be updated and extended for non-deterministic functions.

### Additional context (optional)

so the user can use something like this:
```Python
# not deterministic
Table.shuffle()

# set var to to true after that all calls are deterministic

set_seed = 111

# now seed 111 will be used

# deterministic
Table.shuffle()
```

lars-reimann:
Our functions should always be deterministic since we want the program execution to be reproducible. When we interact with `scikit-learn`, we should thus always specify the `random_seed` parameter of the functions we call (where applicable).

guenterk:
1. Es geht darum, dass ALLES deterministisch ist, typischerweise, um zu testen.
2. Daher ist es am plausibelsten und konsistentesten, statt eines expliziten seeds einen Parameter des SAFE-DS-RUNNERs zu haben, der steuert ob deterministisch oder nicht. Typischerweise ein Boolean.

lars-reimann:
Global state does not work well with memoization and hyperparameter optimization. We'll explicitly pass random seeds into the functions that require it, like `scikit-learn` and `Jax` are doing.

PR #435, opened 2023-07-07
test: `shuffle_rows` should work on `Table`s with multiple rows

### LINKED ISSUE #401: The test for `Table.shuffle_rows` does not test the expected behaviour

The test for `Table.shuffle_rows` does not test wether the rows are shuffled or not as the fiven Table contains only one row. It also tests wether the table and the result are the same after the rows are shuffled.

### Desired solution

Add a test case where the table contains more than one row. Ideally you test that with a given seed the expected order will be given (See the example in the docstrings of this method).
The test **should not** test that the table and the result are the same. They should be different after the rows are shuffled.

_No response_

### Screenshots (optional)

_No response_

_No response_

daniaHu:
Closes #401

### Summary of Changes

Test that `shuffle_rows` also works with a `Table` that contain multiple rows.


alex-senger (approved):

alex-senger (changes requested):
Maybe try adding a `random_state` parameter to the `shuffle_rows` method and putting it into the `random_state` parameter of `_data.sample()`

lars-reimann:
> Maybe try adding a `random_state` parameter to the `shuffle_rows` method and putting it into the `random_state` parameter of `_data.sample()`

I don't think having the `random_state` as a parameter on every method that could be randomized is the way to go (see discussion in #12). It would be possible to just pass a hard-coded `random_state` to `pandas.sample` (e.g. `42` as done in #430) for now to ensure the method is deterministic. But we also discussed last week that it would suffice to test that all rows of the original table are still there after shuffling.

alex-senger (approved):
[LGTM](https://lgtm.party/p/NQ6R7TF8/)

zzril (approved):

zzril (changes requested):
From the PR description:

> Closes #401
> ### Summary of Changes
> 
> Added parameter to test with multiple rows, used np.random.seed to expect the behaviour of shuffle_rows and compared the shuffled table with the expected table

This needs to be updated as you didn't go for the solution using `np.random.seed` in the end.

zzril (approved):

PR #438, opened 2023-07-07
feat: new error class `OutOfBoundsError`

### LINKED ISSUE #262: Custom error if number is out of bounds

zzril:
Closes #262.

### Summary of Changes

* Introduced a new generic error class `OutOfBoundsError` that can be used to signal that a value is outside its expected range.
* Updated Image to use the new error.
* Updated Discretizer to use the new error.
* Updated ML models with hyperparameters to use the new error.


zzril:
Concerning the `NotImplementedError`s: Maybe it's more sensible to use assertions here instead?

@Marsmaennchen221

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,151 @@
...
+    def check_lower_bound(self, value: float) -> bool:
+        """Check that a value is not lower or equal to the Bound."""
+        return value > self._value
+
+    def check_upper_bound(self, value: float) -> bool:
+        """Check that a value is not higher ot equal to the Bound."""
+        return value < self._value
+
+
+class Infinity(OpenBound):
```
lars-reimann, line 116: I don't see the advantage of having `Infinity` and `MinInfinity` classes over only encoding a missing lower/upper bound as `None`. Now there are two ways to express the same thing.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,151 @@
...
+            The actual value that is outside its expected range.
+        lower_bound: Bound | None
+            The lower bound of the expected range.
+        upper_bound: Bound | None
+            The upper bound of the expected range.
+        """
+        if (lower_bound is None or isinstance(lower_bound, MinInfinity)) and (
+            upper_bound is None or isinstance(upper_bound, Infinity)
+        ):
+            raise NotImplementedError("Illegal interval: Value cannot be out of bounds if there are no bounds.")
```
lars-reimann, line 25: `NotImplementedError` should not be used here:

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

https://docs.python.org/3/library/exceptions.html#NotImplementedError

Either raise a `ValueError` or use the [`assert` statement](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

zzril: Changed to `ValueError`.

Thinking more about it, I guess one could even just raise the `OutOfBoundsError` normally, but additionally raise a warning?

src/safeds/exceptions/_generic.py
zzril: Yes, it was mainly to have polymorphism for the stringification.
Made it private and removed it from the tests.

However, the tests also need to test the `check_lower_bound` and `check_upper_bound` methods, so the `_Infinity` and `_MinInfinity` classes need to remain in `__init.py__` in order to stay importable as far as I understand?

zzril:
Something seems wrong with the linter / type checker.
I changed the name of the local variables inside `__init__`, so as to be different from the names of the (differently-typed) arguments, and suddenly things worked.
Looks like the type checker did not understand the fact that there were local variables and incorrectly performed the checks using the types from the arguments instead.

src/safeds/exceptions/_generic.py
lars-reimann: No, you can also directly import declarations from the module they are defined in. The `__init__.py` files are just use to gather declarations from several files.

src/safeds/exceptions/_generic.py
lars-reimann: > Thinking more about it, I guess one could even just raise the `OutOfBoundsError` normally, but additionally raise a warning?

I prefer the value error, becasue it might be hard to link the warning and the out of bounds error together for users.

src/safeds/exceptions/_generic.py
zzril: Done.

src/safeds/exceptions/_generic.py
zzril: Ok, leaving it as `ValueError` then.

I did now, however, include the bounds that were actually given in the error message because the error otherwise seems pretty useless / hard-to-debug.

lars-reimann:
Is it in scope of this PR to also adjust the bounds checks in other places (e.g. `number_of_trees` in `RnadomForest`)? If not, there should be an issue for that.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+        lower_bound: Bound | None
+            The lower bound of the expected range.
+        upper_bound: Bound | None
+            The upper bound of the expected range.
+
+        Raises
+        ------
+        ValueError
+            If upper_bound < lower_bound or if actual does not lie outside the given interval.
+        """
```
lars-reimann, line 26: According to the[ `numpydoc` style guide](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring), parameters of the `__init__` method should be documented in the docstring of the _class_.

It's OK, though, to _repeat_ it in the docstring of the `__init__`, so they also show up in the rendered documentation for `__init__`.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+    def check_upper_bound(self, value: float) -> bool:
+        """Check that a value is not strictly higher than the ClosedBound."""
+        return value <= self._value
+
+
+class OpenBound(Bound):
+    """
+    An open Bound, i.e. the value on the border does not belong to the range.
+
+    May be infinite (unbounded).
```
lars-reimann, line 108: Better remove that since the way to specify that is to omit the bound rather than use the (now internal) bounds for infinity.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+                (
+                    f"Illegal interval: Attempting to raise OutOfBoundsError, but value {actual} is not actually"
+                    f" outside given interval {_lower_bound.str_lower_bound()}, {_upper_bound.str_upper_bound()}."
+                ),
+            )
+        super().__init__(f"{actual} is not inside {_lower_bound.str_lower_bound()}, {_upper_bound.str_upper_bound()}.")
+
+
+class Bound(ABC):
+    """Abstract base class for (lower or upper) Bounds on a float value."""
```
lars-reimann, line 50: `value` parameter is not documented.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+    @abstractmethod
+    def str_lower_bound(self) -> str:
+        """Get a string representation of the Bound as the lower Bound of an interval."""
+
+    @abstractmethod
+    def str_upper_bound(self) -> str:
+        """Get a string representation of the Bound as the upper Bound of an interval."""
+
+    @abstractmethod
+    def check_lower_bound(self, value: float) -> bool:
```
lars-reimann, line 68: `value` parameter is not documented.

Also: `actual` would be a better name, so it matches the constructor of `OutOfBoundsError` and to avoid confusion with `self.value` (same for other places).

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+        """Check that a value does not exceed the Bound on the lower side."""
+
+    @abstractmethod
+    def check_upper_bound(self, value: float) -> bool:
```
lars-reimann, line 72: `value` parameter is not documented.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+        """Check that a value does not exceed the Bound on the upper side."""
+
+    @property
+    def value(self) -> float:
+        """Get the concrete value of the Bound."""
+        return self._value
+
+
+class ClosedBound(Bound):
+    """A closed Bound, i.e. the value on the border belongs to the range."""
```
lars-reimann, line 82: `value` parameter is not documented.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+        return value >= self._value
+
+    def check_upper_bound(self, value: float) -> bool:
+        """Check that a value is not strictly higher than the ClosedBound."""
+        return value <= self._value
+
+
+class OpenBound(Bound):
+    """
+    An open Bound, i.e. the value on the border does not belong to the range.
```
lars-reimann, line 106: `value` parameter is not documented.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+class OutOfBoundsError(ValueError):
+    """A generic exception that can be used to signal that a (float) value is outside its expected range."""
+
+    def __init__(self, actual: float, *, lower_bound: Bound | None = None, upper_bound: Bound | None = None):
+        """
+        Initialize an OutOfBoundsError.
+
+        Parameters
+        ----------
+        actual: float
```
lars-reimann, line 15: Should we treat passing `nan` as `actual` as another (value) error?

zzril:
> Is it in scope of this PR to also adjust the bounds checks in other places (e.g. `number_of_trees` in `RandomForest`)? If not, there should be an issue for that.

I did update a few cases in `_image.py` as there was already a [comment](https://github.com/Safe-DS/Stdlib/issues/262#issuecomment-1614657034) about it in the issue.
Guess I can also check for any other cases where this makes sense and update them.

Converting back to draft meanwhile.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+        """Get a string representation of the concrete value of the Bound."""
+        return "-\u221e"
+
+    def check_lower_bound(self, _: float) -> bool:
+        """Check that a value does not exceed the unrestricted Bound on the lower side. Always true."""
+        return True
+
+    def check_upper_bound(self, _: float) -> bool:
+        """Check that a value does not exceed the unrestricted Bound on the upper side. Always false."""
+        return False
```
lars-reimann, line 166: I'm still not quite happy with these classes:

* `_Infinity` inherits from `OpenBound`. The latter can be used as upper and as lower bound, the former only as an upper bound. Thus, `_Infinity` cannot be used everywhere an `OpenBound` can be used, breaking Liskov's Substitution Principle.
* Similar argument for `_MinInfinity`

I'd rather remove them and adjust the `__str__` methods of `OpenBound` and `ClosedBound` to handle the cases `float("inf")` and `float("-inf")` as you're doing here.

src/safeds/exceptions/_generic.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, value: float):
+        self._value = value
+
+    def __str__(self) -> str:
+        """Get a string representation of the concrete value of the Bound."""
+        return str(self.value)
+
+    @abstractmethod
+    def str_lower_bound(self) -> str:
```
lars-reimann, line 60: These methods (`str_XY`, `check_XY`) should be prefixed with an underscore, since users are not meant to call them.

lars-reimann (changes requested):

lars-reimann:
> > Is it in scope of this PR to also adjust the bounds checks in other places (e.g. `number_of_trees` in `RandomForest`)? If not, there should be an issue for that.
> 
> I did update a few cases in `_image.py` as there was already a [comment](https://github.com/Safe-DS/Stdlib/issues/262#issuecomment-1614657034) about it in the issue. Guess I can also check for any other cases where this makes sense and update them.
> 
> Converting back to draft meanwhile.

Thanks. I think only some hyperparameter of models should be affected. Methods like `Table.get_row` that raise an `IndexOutOfBoundsError` can be kept unchanged.

src/safeds/exceptions/_generic.py
lars-reimann: This check could also be added to `Bound.__init__`.

src/safeds/exceptions/_generic.py
zzril: Makes sense, did that now.

src/safeds/exceptions/_generic.py
zzril: Done.

src/safeds/exceptions/_generic.py
zzril: I can't think of a situation where this would make sense, so I added a check in `Bound.__init__`, raising a `ValueError` if `nan` is given.
If the need to pass `nan` ever occurs, we can allow it then.

src/safeds/exceptions/_generic.py
zzril: Done.

src/safeds/exceptions/_generic.py
zzril: Removed that.
However, the docstring of `Bound.__init__(self, value)` now states that *some* bounds *may* take allow `+/-inf` as a parameter.
I could also "forbid" it explicitely in the docstring for `Bound`, then still allow it in `OpenBound`, but that somehow feels incorrect.

src/safeds/exceptions/_generic.py
zzril: Done.

src/safeds/exceptions/_generic.py
zzril: Makes sense. Done.

src/safeds/exceptions/_generic.py
zzril: Done.

src/safeds/exceptions/_generic.py
zzril: Done.

src/safeds/exceptions/_generic.py
zzril: Done.

zzril:
Added a `name` parameter to the `OutOfBoundsError` because otherwise, the error messages would have become less useful.

src/safeds/exceptions/_generic.py
lars-reimann: Makes sense. Then we should also check `actual` in `OutOfBoundsError.__init__`.

src/safeds/exceptions/_generic.py
lars-reimann: It's OK how you've done it. Alternatively, we could omit the discussion of infinity altogether for the bound classes since it's just meant for our internal use in the same file.

zzril: I have changed the design to still allow creating `Bound`s with `float("inf")` (as we need to create these bounds internally), but to not allow passing such bounds to the constructor of the `OutOfBoundsError`. Upon trying, there will be a `ValueError` stating explicitely that `None` should be used.

src/safeds/exceptions/_generic.py
zzril: Added checks for that.

zzril:
In [_k_nearest_neighbors.py](https://github.com/Safe-DS/Stdlib/blob/2c54ec0ddf269eabe9052a526f72b9446a8b0d02/src/safeds/ml/classical/classification/_k_nearest_neighbors.py#L92-L98):
```py
        if self._number_of_neighbors > training_set.number_of_rows:
            raise ValueError(
                (
                    f"The parameter 'number_of_neighbors' ({self._number_of_neighbors}) has to be less than or equal to"
                    f" the sample size ({training_set.number_of_rows})."
                ),
            )
```

We can replace this with the new `OutOfBoundsError`, but then we cannot customize the error message as much. (It does allow for a `name` keyword argument, so we can tell the user the name of the `number_of_neighbors` parameter, but we cannot talk about the "sample size" anymore.)

We could make the `OutOfBoundsError` more flexible to also allow passing arbitrary strings to it (for example with a new `custom_message: str` keyword argument).
But then the error itself gets somewhat complicated to use...

zzril:
> > > Is it in scope of this PR to also adjust the bounds checks in other places (e.g. `number_of_trees` in `RandomForest`)? If not, there should be an issue for that.
> > 
> > 
> > I did update a few cases in `_image.py` as there was already a [comment](https://github.com/Safe-DS/Stdlib/issues/262#issuecomment-1614657034) about it in the issue. Guess I can also check for any other cases where this makes sense and update them.
> > Converting back to draft meanwhile.
> 
> Thanks. I think only some hyperparameter of models should be affected. Methods like `Table.get_row` that raise an `IndexOutOfBoundsError` can be kept unchanged.
> 
> Edit: An the `number_of_bins` on `Discretizer`.

Changed all ML models (and the discretizer) to use the new error.

lars-reimann:
> In [_k_nearest_neighbors.py](https://github.com/Safe-DS/Stdlib/blob/2c54ec0ddf269eabe9052a526f72b9446a8b0d02/src/safeds/ml/classical/classification/_k_nearest_neighbors.py#L92-L98):
> 
> ```python
>         if self._number_of_neighbors > training_set.number_of_rows:
>             raise ValueError(
>                 (
>                     f"The parameter 'number_of_neighbors' ({self._number_of_neighbors}) has to be less than or equal to"
>                     f" the sample size ({training_set.number_of_rows})."
>                 ),
>             )
> ```
> 
> (Same in the `regression` version.)
> 
> We can replace this with the new `OutOfBoundsError`, but then we cannot customize the error message as much. (It does allow for a `name` keyword argument, so we can tell the user the name of the `number_of_neighbors` parameter, but we cannot talk about the "sample size" anymore.)
> 
> We could make the `OutOfBoundsError` more flexible to also allow passing arbitrary strings to it (for example with a new `custom_message: str` keyword argument). But then the error itself gets somewhat complicated to use...

Let's keep this one as is. `OutOfBoundsError` makes sense if the bounds are some fixed constant rather than something like the length of a list.

lars-reimann (approved):
Thanks a lot, very nice work.

PR #446, opened 2023-07-11
docs: Added maximum height to output execution cells in mkdocs to prevent large tables to fill the whole site

### LINKED ISSUE #447: Data processing tutorial loads very slowly
The data processing tutorial loads very slowly since #358 due to its sheer size (~3.5MB). We need to truncate its output.

Marsmaennchen221:
Closes #447.

### Summary of Changes

docs: Added maximum height to output execution cells in mkdocs to prevent large tables to fill the whole site
docs: Removed a warning in the data processing tutorial
docs: Added table of contents to data processing tutorial
docs: Shorten output in "Data Processing" tutorial to speed up loading the page

Used the html note like we discussed in [#419](https://github.com/Safe-DS/Stdlib/pull/419#discussion_r1259709264)


lars-reimann:
Good idea and I'd definitely keep the max height. However, we also need to keep in mind that the data will still bloat the generated HTML, so truncating the output is still required to keep its size down.

There's also still one warning:

image

zzril:
> docs: Added table of contents to data processing tutorial

There should be more chapters. At the very least, a "General" section for everything before the `Transform Table`/`Colunm` sections.
Possible use a separate section for loading the data as well.

Also, now with #419 merged, that one should probably also get such a table of content now.

lars-reimann (approved):
There's now another warning:

image

I don't find that helpful though since the example code is perfectly valid and will remove it in another PR.

PR #450, opened 2023-07-13
feat: Improve error handling of TaggedTable

### LINKED ISSUE #150: Improve error handling of `TaggedTable`

jxnior01:
Closes #150 

### Summary of Changes

* feat: Validated inputs of functions
* feat: Raised appropriate exceptions with appropriate messages
* docs: Modified docs
* tests: Tested all exceptions


Marsmaennchen221:
Do we really want to catch every error twice? We already have the logic in the `Table` methods so the logic stays the same. If we want to change the logic of one of the errors we would have to do this in both the `Table` and the `TaggedTable`. I would prefer to only write additional errors in `TaggedTable` if they only occur in the `TaggedTable` methods or if they differ from the `Table` error logic/messages

jxnior01:
> Do we really want to catch every error twice? We already have the logic in the `Table` methods so the logic stays the same. If we want to change the logic of one of the errors we would have to do this in both the `Table` and the `TaggedTable`. I would prefer to only write additional errors in `TaggedTable` if they only occur in the `TaggedTable` methods or if they differ from the `Table` error logic/messages

That makes sense, it's not ideal. I'll revise my changes to include additional error handling where necessary.

Marsmaennchen221:
@jxnior01 I removed additional errors that are being raised by the methods from the `Table` class

jxnior01:
> @jxnior01 I removed additional errors that are being raised by the methods from the `Table` class

Thanks. That was helpful.

Marsmaennchen221 (approved):

zzril:
Looks good to me, although I've spotted a few cases where we don't make use of the new "suggest columns" feature. Not sure what our criterion is about when to make use of that.

Can merge if you like.

PR #451, opened 2023-07-13
feat: Added `Table.list_columns_with_missing_values`

### LINKED ISSUE #44: Inferred schema should know which columns have missing values

Marsmaennchen221:
Closes #44

### Summary of Changes

feat: Added `Table.list_columns_with_missing_values`


Marsmaennchen221:
I dont think the other changed requested in #44 make sense anymore. The requested `OptionalColumnType` is now implemented with the nullable types. The logic that converts "normal" columns into "nullable" columns when adding a missing value is implemented. 
I would leave the method `Column.has_missing_values` as it is now because with the `ColumnType` we only know that the column **could** contain missing values not that it has those in its current state.
If wanted we can add a different method `Column.could_contain_missing_values` but I am not sure if this is needed.

@lars-reimann Your thoughts?

lars-reimann:
> I dont think the other changed requested in #44 make sense anymore. The requested `OptionalColumnType` is now implemented with the nullable types. The logic that converts "normal" columns into "nullable" columns when adding a missing value is implemented. I would leave the method `Column.has_missing_values` as it is now because with the `ColumnType` we only know that the column **could** contain missing values not that it has those in its current state. If wanted we can add a different method `Column.could_contain_missing_values` but I am not sure if this is needed.
> 
> @lars-reimann Your thoughts?

Let's move this discussion to #44.

jxnior01 (approved):
LGTM :D

Marsmaennchen221:
> > I dont think the other changed requested in #44 make sense anymore. The requested `OptionalColumnType` is now implemented with the nullable types. The logic that converts "normal" columns into "nullable" columns when adding a missing value is implemented. I would leave the method `Column.has_missing_values` as it is now because with the `ColumnType` we only know that the column **could** contain missing values not that it has those in its current state. If wanted we can add a different method `Column.could_contain_missing_values` but I am not sure if this is needed.
> > @lars-reimann Your thoughts?
> 
> Let's move this discussion to #44.

I will leave this pr as a draft for now until we have solved this discussion

lars-reimann:
Closing this, because the main branch has heavily diverged from this one.

PR #530, opened 2024-01-18
feat: Add adjust_color_balance method in Image

### LINKED ISSUE #525: Add adjust_color_balance method in Image

The Image Class prior to #506 had an adjust_color_balance method, which we removed because PyTorch has no such method.

### Desired solution

Implement a method with that functionality, or implement a pre-trained NN.

_No response_

### Screenshots (optional)

_No response_

_No response_

Marsmaennchen221:
Closes #525

### Summary of Changes

Added adjust_color_balance method for the new image class. Works just like the adjust_color_balance method from the old image with pillow.


src/safeds/data/image/containers/_image.py
```diff
@@ -434,6 +430,38 @@ def adjust_contrast(self, factor: float) -> Image:
...
+
+        The original image is not modified.
+
+        Parameters
+        ----------
+        factor: float
+            If factor > 1, increase color balance of image.
+            If factor = 1, no changes will be made.
+            If factor < 1, make image greyer.
+            Has to be bigger than or equal to 0.
```
lars-reimann, line 445: I'd suggest changing the order of those entries, so the intervals are listed from low to high:

```suggestion
            Has to be bigger than or equal to 0.
            If 0 <= factor < 1, make image greyer.
            If factor = 1, no changes will be made.
            If factor > 1, increase color balance of image.
```

lars-reimann (approved):

PR #547, opened 2024-02-05
feat: return the correct size for custom container objects

WinPlay02:
This is needed to correctly evaluate whether an object is worth memoizing or keeping in the cache.

See for context: https://github.com/Safe-DS/Runner/pull/51 and https://github.com/Safe-DS/Runner/issues/44

For future container classes (like e.g. image set this would also need to be added, to be compatible with the memoizing implementation in the runner)

### DIFF
--- a/src/safeds/data/image/containers/_image.py
+++ b/src/safeds/data/image/containers/_image.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 import io
 import warnings
 from pathlib import Path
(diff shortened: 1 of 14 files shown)


lars-reimann (changes requested):
Looks great. I only have a small request: Please move the new `__sizeof__` methods towards the beginning of the classes, where the other dunder methods (e.g. `__eq__` and `__hash__` are listed). That makes it easier to find them in the rather large classes. 

It's not that important how the dunder methods themselves are sorted, but if you want to follow the pattern used previously, list `__init__` first and sort the rest alphabetically.

PR #550, opened 2024-02-21
feat: added normal plot for time series

### LINKED ISSUE #549: feat: normal visualization of time series

Just a normal plot for the time series

### Desired solution

```TimeSeries.plot_time_series()``` function

### PR DESCRIPTION
**Pull request by Gerhardsa0**
Closes #549 

### Summary of Changes

Add more plot for time series:

* ```TimeSeries.plot_lineplot```
* ```TimeSeries.plot_scatterplot```

### CODE REVIEW

**Comment by lars-reimann** [comment]
Wouldn't it also make sense to plot numeric feature columns against the time column? In that case, this function should get a parameter `column_name` to select the column to plot.

**Comment by Gerhardsa0** [comment]
> Wouldn't it also make sense to plot numeric feature columns against the time column? In that case, this function should get a parameter `column_name` to select the column to plot.

Yes, it would give a user a more detailed view on, how the other data points are correlated to the target column. On the other hand, the plots can get messy, but only if a lot of columns are plotted. Also, it should be an optional argument, so the user can always easy plot the time series.

**Comment by lars-reimann** [comment]
> On the other hand, the plots can get messy, but only if a lot of columns are plotted.

I was only thinking of plotting _one_ feature column against the time column.

> Yes, it would give a user a more detailed view on, how the other data points are correlated to the target column. Also, it should be an optional argument, so the user can always easy plot the time series.

Then let's add an optional `column_name` parameter. If it's not set, we plot the target column. If it is set, we plot the selected column. Validation that the column is numeric should happen in any case.

**Comment by Gerhardsa0** [comment]
Should a user choose between scatter and normal plot?

**Comment by lars-reimann** [comment]
> Should a user choose between scatter and normal plot?

In the `Table` class we have [`plot_lineplot`](https://library.safeds.com/en/stable/reference/safeds/data/tabular/containers/Table/#safeds.data.tabular.containers.Table.plot_lineplot) and [`plot_scatterplot`](https://library.safeds.com/en/stable/reference/safeds/data/tabular/containers/Table/#safeds.data.tabular.containers.Table.plot_scatterplot). Both take `x_column_name` and a `y_column_name` as _required_ parameters.

We could also have those two methods in `TimeSeries`, but add `x_column_name` and `y_column_name` as _optional_ parameters. The default `x_column_name` would be the name of the time column and the default `y_column_name` would be the name of the feature column.

**Comment by Gerhardsa0** [comment]
> > Should a user choose between scatter and normal plot?
> 
> In the `Table` class we have [`plot_lineplot`](https://library.safeds.com/en/stable/reference/safeds/data/tabular/containers/Table/#safeds.data.tabular.containers.Table.plot_lineplot) and [`plot_scatterplot`](https://library.safeds.com/en/stable/reference/safeds/data/tabular/containers/Table/#safeds.data.tabular.containers.Table.plot_scatterplot). Both take `x_column_name` and a `y_column_name` as _required_ parameters.
> 
> We could also have those two methods in `TimeSeries`, but make `x_column_name` and `y_column_name` _optional_ parameters. The default `x_column_name` would be the name of the time column and the default `y_column_name` would be the name of the feature column.

With the `y_column_name` I agree, if the user wants a different name on his plot, for whatever reason. But with the `x_column_name` the time series mostly  isthe given date column here, so this should be the x axis. For that reason we could also add an optional date column parameter for time series, so the plots look better.

**Comment by Gerhardsa0** [comment]
> > Should a user choose between scatter and normal plot?
> 
> In the `Table` class we have [`plot_lineplot`](https://library.safeds.com/en/stable/reference/safeds/data/tabular/containers/Table/#safeds.data.tabular.containers.Table.plot_lineplot) and [`plot_scatterplot`](https://library.safeds.com/en/stable/reference/safeds/data/tabular/containers/Table/#safeds.data.tabular.containers.Table.plot_scatterplot). Both take `x_column_name` and a `y_column_name` as _required_ parameters.
> 
> We could also have those two methods in `TimeSeries`, but make `x_column_name` and `y_column_name` _optional_ parameters. The default `x_column_name` would be the name of the time column and the default `y_column_name` would be the name of the feature column.

The X-Axis is usally the date in time series or the usual index column. Should I still add the `x_column_name` ? and should the time series take an optional parameter for a date column?

**Comment by lars-reimann** [comment]
> The X-Axis is usally the date in time series or the usual index column. Should I still add the `x_column_name` ?

I'd say yes due to two reasons:
1. Symmetry with the `Table` class.
2. It's useful in order to see correlation between two feature columns.

> and should the time series take an optional parameter for a date column?

Hmm, probably not. The time column should include dates (timestamps). Otherwise, we would also need to support separate columns for year/month/day, hour/minute/second/nanos or combinations thereof.

File: src/safeds/data/tabular/containers/_time_series.py [hunk 1]
```diff
@@ -906,3 +907,151 @@ def plot_lagplot(self, lag: int) -> Image:
         plt.close()  # Prevents the figure from being displayed directly
         buffer.seek(0)
         return Image.from_bytes(buffer.read())
+
+    def plot_time_series_lineplot(self, y_column_name: str | None = None, x_column_name: str | None = None) -> Image:
```
lars-reimann, line 911:
```suggestion
    def plot_lineplot(self, x_column_name: str | None = None, y_column_name: str | None = None) -> Image:
```

* We already know that we are calling the method on a time series, so there's no reason to repeat that. 
* Specifying x first is more natural. It's also consistent with `Table`

src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +907,151 @@ def plot_lagplot(self, lag: int) -> Image:
...
+
+        Examples
+        --------
+                >>> from safeds.data.tabular.containers import TimeSeries
+                >>> table = TimeSeries({"time":[1, 2], "target": [3, 4], "feature":[2,2]}, target_name= "target", time_name="time", feature_names=["feature"], )
+                >>> image = table.plot_time_series_lineplot()
+
+        """
+        self._data.index.name = "index"
+        # falls mitgegebene column kein numerical column ist
```
lars-reimann, line 947: Should be in English.

src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +907,151 @@ def plot_lagplot(self, lag: int) -> Image:
...
+        plt.tight_layout()
+
+        buffer = io.BytesIO()
+        fig.savefig(buffer, format="png")
+        plt.close()  # Prevents the figure from being displayed directly
+        buffer.seek(0)
+        self._data = self._data.reset_index()
+        return Image.from_bytes(buffer.read())
+
+    def plot_time_series_scatterplot(
```
lars-reimann, line 985:
```suggestion
    def plot_scatterplot(
```

src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +907,151 @@ def plot_lagplot(self, lag: int) -> Image:
...
+    ) -> Image:
+        """
+        Plot the time series target or the given column(s) as scatter plot.
+
+        The function will take the target column as the default value for x_column_name and a time dummy column as the
+        default value for y_column_name.
+
+        Parameters
+        ----------
+        x_column_name : str
```
lars-reimann, line 998:
```suggestion
        x_column_name
```

src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +907,151 @@ def plot_lagplot(self, lag: int) -> Image:
...
+            The column name of the column to be plotted on the x-Axis.
+        y_column_name : str
```
lars-reimann, line 1000:
```suggestion
        y_column_name
```

tests/safeds/data/tabular/containers/_table/_tagged_table/_time_series/test_plot_time_series.py
lars-reimann, line 1: Rename this file accordingly to `test_plot_lineplot`. The location should also mirror the src directory, so move this to `tests/.../containers/_time_series`

tests/safeds/data/tabular/containers/_table/_tagged_table/_time_series/test_plot_time_series_scatter.py
lars-reimann, line 1: Rename this file accordingly to `test_plot_scatterplot`. The location should also mirror the src directory, so move this to `tests/.../containers/_time_series`

lars-reimann (changes requested):

lars-reimann:
> The default value of, `y_column_name ` should be the index, so the user can simple plot the timeseries with a call like: `TimeSeries.plot_time_series()`. Or is this not wanted behavior?

Would the time column not make more sense?

Gerhardsa0:
> > The default value of, `y_column_name ` should be the index, so the user can simple plot the timeseries with a call like: `TimeSeries.plot_time_series()`. Or is this not wanted behavior?
> 
> Would the **time column** not make more sense as the default **x** `_column_name`? The default **y** `_column_name` should be the target.

Yes, I didn't notice.

src/safeds/data/tabular/containers/_time_series.py
Gerhardsa0: changed

src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +907,151 @@ def plot_lagplot(self, lag: int) -> Image:
...
+        """
+
+        Plot the time series target or the given column(s) as line plot.
+
+        The function will take the target column as the default value for x_column_name and a time dummy column as the
+        default value for y_column_name.
+
+        Parameters
+        ----------
+        x_column_name : str
```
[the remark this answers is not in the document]
Gerhardsa0: deleted

src/safeds/data/tabular/containers/_time_series.py
Gerhardsa0: deleted

src/safeds/data/tabular/containers/_time_series.py
Gerhardsa0: deleted

tests/safeds/data/tabular/containers/_table/_tagged_table/_time_series/test_plot_time_series.py
Gerhardsa0: changed

Gerhardsa0:
> > The default value of, `y_column_name ` should be the index, so the user can simple plot the timeseries with a call like: `TimeSeries.plot_time_series()`. Or is this not wanted behavior?
> 
> Would the **time column** not make more sense as the default **x** `_column_name`? The default **y** `_column_name` should be the target.

I added this behavior, to the function, which throws up new questions.

The time column can be pretty useful for plotting, so the user could see dates in the plot. But we don't handle them as dates, only as a time dummy right now.

Should the time column only allow numerical values? As we don't want it to handle, “real” date-types.

lars-reimann (approved):
From my point of view, this is ready to merge. Thank you!

PR #551, opened 2024-02-21
feat: renamed plot_lagplot -> plot_lag_plot

### LINKED ISSUE #521: feat: Add moving average plot

The user should be able to see the moving average in context with the time series.

### Desired solution

The user should just call something like this: ```ts.plot_movingaverage()```
When implementing this, the pandas ```df.roll```-function could be pretty useful.

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html

Gerhardsa0:
Closes #521 

### Summary of Changes
added moving average plot,


src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +906,64 @@ def plot_lagplot(self, lag: int) -> Image:
         plt.close()  # Prevents the figure from being displayed directly
         buffer.seek(0)
         return Image.from_bytes(buffer.read())
+
+    def plot_moving_average(
+        self,
+        window_size: int,
+        feature_name: str | None = None,
```
lars-reimann, line 913: Rename to `column_name` (could also explicitly be the target).

src/safeds/data/tabular/containers/_time_series.py
```diff
@@ -906,3 +906,64 @@ def plot_lagplot(self, lag: int) -> Image:
...
+        ax_temp = series_mvg.plot()
+        ax = series.plot(ax=ax_temp)
+        ax.legend(labels=["moving_average", feature_name])
+
+        fig = ax.figure
+        buffer = io.BytesIO()
+        fig.savefig(buffer, format="png")
+        plt.close()  # Prevents the figure from being displayed directly
+        buffer.seek(0)
+        return Image.from_bytes(buffer.read())
```
lars-reimann, line 969: We are using this for all plots. Can we extract that into an internal function`_create_image_for_plot(fig: ?) -> Image` in a file `src/safeds/_utils/_plotting.py`? Also add a reexport of the function to `src/safeds/_utils/__init__.py`.

lars-reimann (changes requested):

src/safeds/data/tabular/containers/_time_series.py
Gerhardsa0: added

PR #573, opened 2024-03-15
feat: add deterministic hash methods to all types

WinPlay02:
### Summary of Changes

- feat: add deterministic `__hash__` methods to the all types
- test: add tests to all new hash implementations
- feat: add `__eq__` method to `TaggedTable` and `TimeSeries`

The `__hash__`-Implementation is needed to more efficiently check whether data may be the same.
A deterministic implementation is used (based on `xxhash`, a fast non-cryptographic hash algorithm) to allow these comparisons to be performed across different interpreters (processes).

The implementation of the Schema was updated to be deterministic.


lars-reimann (changes requested):

lars-reimann (approved):
Looks great.

Issue #575, opened 2024-03-17
performance problem: various table operations


WinPlay02:
### Describe the bug

Some `Table` operations take multiple hours to complete on large tables, although they should be much faster. When using pandas `loc` functionality, these operations complete in a few seconds.
These operations include:
- [ ] `filter_rows`
- [ ] `group_rows_by`
- [x] `add_rows`(Table)
- [ ] `sort_rows` (probably, untested)
- [ ] `transform_column` (probably, untested)

The common problem of these operations are, that the first step is boxing the rows to a pandas `Series` and wrapping them in a `Row` object (which both leads to a huge overhead).
Most of the time, the single row objects are only needed to be passed to the lambda function that is provided to these operations.

### Possible resolutions:
- don't guarantee that a `Row` is passed to the lambda functions (for `filter_rows`, `group_rows_by`, `sort_rows`, `transform_column`). Instead only pass a general type that is indexable with Strings (column names), like `Mapping[str, Any]` to access the values. Maybe also pass the schema, as the other needed operations on a `Row` object can be handled with a schema.
- `add_rows` could handle adding a whole table as a special case, so they are not split into `Rows` first, avoiding the problem without changing the interface

### To Reproduce

1. Load a huge dataset with `Table`.`from_csv_file`
2. Run any of the previously mentioned operations
3. Observe that nothing happens and after a few hours the pipeline still runs

### Expected behavior

Faster execution of the listed operations

WinPlay02:
Concrete solution for `Table.add_rows` (insert at the top of the method):
```python
        if isinstance(rows, Table):
            if rows.number_of_rows == 0:
                return self
            if self.number_of_columns == 0:
                return rows
            different_column_names = set()
            different_column_names.update(set(self.column_names) - set(rows.column_names))
            if len(different_column_names) > 0:
                raise UnknownColumnNameError(
                    sorted(
                        different_column_names,
                        key={val: ix for ix, val in enumerate(self.column_names)}.__getitem__,
                    ),
                )

            new_df = pd.concat([self._data, rows._data]).infer_objects()
            new_df.columns = self.column_names
            schema = Schema.merge_multiple_schemas([self.schema, rows.schema])
            result = Table._from_pandas_dataframe(new_df, schema)

            return result
```

lars-reimann:
Fixed in #734 and #738.

PR #606, opened 2024-04-03
perf: treat Tables specially when calling add_rows

WinPlay02:
Fixes partially #575

### Summary of Changes

Add special case to `add_rows` when adding a `Table`. 
The table is no longer split into a list of rows, instead it is directly added to the `DataFrame`.
In special cases, this may improve the performance up to 1000x.

### DIFF
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -1074,9 +1074,7 @@ class Table:
             new_df = pd.concat([self._data, rows._data]).infer_objects()
             new_df.columns = self.column_names
             schema = Schema._merge_multiple_schemas([self.schema, rows.schema])
-            result = Table._from_pandas_dataframe(new_df, schema)
-
-            return result
+            return Table._from_pandas_dataframe(new_df, schema)
 
         if len(rows) == 0:
             return self
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -1057,7 +1057,26 @@ class Table:
         2  5  6
         """
         if isinstance(rows, Table):
-            rows = rows.to_rows()
+            if rows.number_of_rows == 0:
+                return self
+            if self.number_of_columns == 0:
+                return rows
+            different_column_names = set()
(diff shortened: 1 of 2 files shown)


src/safeds/data/tabular/containers/_table.py
```diff
@@ -1057,7 +1057,24 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
         2  5  6
         """
         if isinstance(rows, Table):
-            rows = rows.to_rows()
+            if rows.number_of_rows == 0:
+                return self
+            if self.number_of_columns == 0:
+                return rows
+            different_column_names = set()
+            different_column_names.update(set(self.column_names) - set(rows.column_names))
```
lars-reimann, line 1065:
```suggestion
            different_column_names = set(self.column_names) - set(rows.column_names)
```

lars-reimann (approved):
Thanks, nice contribution!

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1057,7 +1057,23 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
...
+                return rows
+            different_column_names = set(self.column_names) - set(rows.column_names)
+            if len(different_column_names) > 0:
+                raise UnknownColumnNameError(
+                    sorted(
+                        different_column_names,
+                        key={val: ix for ix, val in enumerate(self.column_names)}.__getitem__,
+                    ),
+                )
+
```
Marsmaennchen221, line 1072:
```suggestion
            if self.number_of_rows == 0:
                return rows
            
```

Marsmaennchen221: Well...
Seems like I was a bit too late :|

WinPlay02: @Marsmaennchen221 I can create a follow-up PR if you want

lars-reimann: Good point, though. Is this about line 1062 (`if self.number_of_columns == 0`?

WinPlay02: I'd check for either, as in either case it should be valid to do nothing

PR #607, opened 2024-04-03
perf: add_rows follow-up, also check for empty table (no rows)

WinPlay02:
See #606

- another way to check for empty tables (no rows)

### DIFF
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -1057,7 +1057,7 @@ class Table:
         2  5  6
         """
         if isinstance(rows, Table):
-            if rows.number_of_rows == 0:
+            if rows.number_of_columns == 0 or rows.number_of_rows == 0:
                 return self
             if self.number_of_columns == 0 or self.number_of_rows == 0:
                 return rows
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -1059,7 +1059,7 @@ class Table:
         if isinstance(rows, Table):
             if rows.number_of_rows == 0:
                 return self
-            if self.number_of_columns == 0:
+            if self.number_of_columns == 0 or self.number_of_rows == 0:
                 return rows
             different_column_names = set(self.column_names) - set(rows.column_names)
             if len(different_column_names) > 0:


WinPlay02:
@Marsmaennchen221 Here is the check for amount of rows equals 0

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1059,7 +1059,7 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
         if isinstance(rows, Table):
             if rows.number_of_rows == 0:
```
lars-reimann, line 1060: I guess if we check `number_of_columns` in the next condition, we should also check it here:

```suggestion
            if rows.number_of_columns == 0 or rows.number_of_rows == 0:
```

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1057,9 +1057,9 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
         2  5  6
         """
         if isinstance(rows, Table):
-            if rows.number_of_rows == 0:
+            if rows.number_of_columns == 0 or rows.number_of_rows == 0:
```
Marsmaennchen221, line 1060: The second check is not really necessary since number_of_colums == 0 => number_of_rows == 0

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1057,9 +1057,9 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
...
                 return self
-            if self.number_of_columns == 0:
+            if self.number_of_columns == 0 or self.number_of_rows == 0:
```
Marsmaennchen221, line 1062: We should check after comparing columns to check for different schemas

Marsmaennchen221:
See #608

src/safeds/data/tabular/containers/_table.py
WinPlay02: @Marsmaennchen221 But is the schema relevant, if there is nothing changed either way?

src/safeds/data/tabular/containers/_table.py
```diff
@@ -1059,7 +1059,7 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
         if isinstance(rows, Table):
             if rows.number_of_rows == 0:
                 return self
-            if self.number_of_columns == 0:
+            if self.number_of_columns == 0 or self.number_of_rows == 0:
                 return rows
             different_column_names = set(self.column_names) - set(rows.column_names)
```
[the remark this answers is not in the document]
WinPlay02: Symmetric difference seems correct

WinPlay02:
Superseded by #608

PR #608, opened 2024-04-03
perf: Add special case to `Table.add_rows` to increase performance

Marsmaennchen221:
See #606 

### Summary of Changes

perf: Add special case to `Table.add_rows` to increase performance

### DIFF
--- a/src/safeds/data/tabular/containers/_table.py
+++ b/src/safeds/data/tabular/containers/_table.py
@@ -1069,6 +1069,8 @@ class Table:
                         key={val: ix for ix, val in enumerate(self.column_names)}.__getitem__,
                     ),
                 )
+            if self.number_of_rows == 0:
+                return rows
 
             new_df = pd.concat([self._data, rows._data]).infer_objects()
             new_df.columns = self.column_names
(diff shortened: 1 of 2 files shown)


src/safeds/data/tabular/containers/_table.py
```diff
@@ -1069,6 +1069,8 @@ def add_rows(self, rows: list[Row] | Table) -> Table:
                         key={val: ix for ix, val in enumerate(self.column_names)}.__getitem__,
                     ),
                 )
+            if self.number_of_rows == 0:
```
WinPlay02, line 1072:
```suggestion
            if rows.number_of_rows == 0:
                return self
            if self.number_of_rows == 0:
```

Then we also don't concat if there's nothing to do, the other way round

WinPlay02 (approved):
Consider the suggestion, otherwise I agree. I'll close my second PR then.

I guess checking the schema will ensure correctness and will prevent surprising the user, that sometimes an `add_rows` operation works, and sometimes it doesn't

WinPlay02 (approved):
Looks good

lars-reimann (approved):

PR #609, opened 2024-04-04
feat: more hash, sizeof and eq implementations

WinPlay02:
### Summary of Changes

- adds specific hash-implementations to classical classification and regression models
- use more properties from table transformer to calculate hash
- add eq to imputer strategies and svm kernels
- adds general reusable structural hashing infrastructure for future usage


WinPlay02:
@lars-reimann The linter seems broken, it doesn't automatically push anything anymore

lars-reimann:
> @lars-reimann The linter seems broken, it doesn't automatically push anything anymore

Some errors cannot be fixed automatically. If there is at least one such error, no commit is created to fix the others. I'll look into why it creates no comment, though.

WinPlay02:
> > @lars-reimann The linter seems broken, it doesn't automatically push anything anymore
> 
> Some errors cannot be fixed automatically. If there is at least one such error, no commit is created to fix the others. I'll look into whether it can create comments on PRs from forks, though.

See this commit: https://github.com/Safe-DS/Library/pull/609/commits/68c3562817265a96de0f52628458cc020c88add9 (https://github.com/Safe-DS/Library/actions/runs/8553108748/job/23435588119). No errors are left, and the linter still does not push (although at least 22 fixes could be made). Maybe this is also a problem with the fork.

lars-reimann:
> Maybe this is also a problem with the fork.

Yep, it is. Workflows with the `pull_request` trigger can be changed by forks, and thus only run with read permissions for security.

src/safeds/data/tabular/transformation/_table_transformer.py
```diff
@@ -21,7 +23,16 @@ def __hash__(self) -> int:
...
+        added = self.get_names_of_added_columns() if self.is_fitted() else []
+        changed = self.get_names_of_changed_columns() if self.is_fitted() else []
+        removed = self.get_names_of_removed_columns() if self.is_fitted() else []
+        return xxhash.xxh3_64(
+            self.__class__.__qualname__.encode("utf-8")
+            + (1 if self.is_fitted() else 0).to_bytes(1)
+            + functools.reduce(operator.add, [feature.encode("utf-8") for feature in added], b"A")
+            + functools.reduce(operator.add, [feature.encode("utf-8") for feature in changed], b"C")
+            + functools.reduce(operator.add, [feature.encode("utf-8") for feature in removed], b"R"),
+        ).intdigest()
```
lars-reimann, line 35: It seems like your pattern for creating a hash is always the same:

1. `None` -> `b"\0"`
2. `int` -> `to_bytes`
3. `bool` -> `to 0/1`, then `to_bytes`
4. `string` -> `encode`
5. `float` -> `struct` pack
6. `list` (or some other iterable) -> `reduce` + `add` $^1$
7. other -> `hash`

Thus, I'd strongly suggest capturing this logic in an extracted method, so it's easier to add hashing for other conceptually immutable classes:

* Create a `_utils` package in `src/safeds` next to `_config`.
* Add a module `_hashing.py` in there.
* Add a function `hash(*value: any) -> int` in there

$^1$ The different initial values for `reduce` (here `b"A"`) could either be omitted or we keep track of a counter when we hash the arguments passed to `hash`.

lars-reimann (changes requested):

src/safeds/data/tabular/transformation/_imputer.py
```diff
@@ -47,9 +49,29 @@ class Constant(ImputerStrategy):
...
 
+            def __eq__(self, other: object) -> bool:
+                if not isinstance(other, Imputer.Strategy.Constant):
+                    return NotImplemented
+                if self is other:
+                    return True
+                return self._value == other._value
+
+            def __hash__(self) -> int:
+                return xxhash.xxh3_64(self.__class__.__qualname__.encode("utf-8")).intdigest()
```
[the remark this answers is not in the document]
WinPlay02: As it is `Any` value, it cannot be part of the hash (no determinism guarantee).

Pulling it up into the superclass `ImputerStrategy` didn't work for me, as `__eq__` is overridden here, which deletes the `__hash__` from the superclass... (also applies to the other comments)

src/safeds/data/tabular/transformation/_table_transformer.py
WinPlay02: Will do :+1:

lars-reimann: Maybe `structuralHash` would be fitting names (like [structural equality](https://kotlinlang.org/docs/equality.html#structural-equality) vs [referential equality](https://kotlinlang.org/docs/equality.html#referential-equality)).

WinPlay02:
Everything should be addressed now, I even learned that implementations of methods can be copied from the parent class (for deduplicating `__hash__` implementations)

lars-reimann (approved):
The hash functions are very nice to read now. All looks good.

PR #616, opened 2024-04-09
perf: improve performance of model & forward layer

### LINKED ISSUE #610: Improve FNN Layer and Model Performance

Currently our model is doing fine performancewise, since we just support FNN-Layers. But since we will also implement CNNs, which partly depend on FNN-Layers it is important for them to be performance-optimal, since small suboptimal implementations that don't really impact FNNs, will have a huge impact on computing time when working with Images.

### Desired solution

Rework some code parts to optimize computing time and fix some bugs, everything discussed with @Marsmaennchen221 

_No response_

### Screenshots (optional)

_No response_

_No response_

sibre28:
Closes #610 

### Summary of Changes

Fixed some bugs and improved the performance of some methods, there are still some changes to be made but it is helpful to merge this now as @Marsmaennchen221 and @Gerhardsa0 partly depend on it


WinPlay02 (commented):
Just a friendly reminder (as I do not know about the planing in the *Library* group): Please add `__hash__`, `__eq__` and `__sizeof__` implementations where possible (as they are really valuable for running the library with the memoization infrastructure provided by the runner)

E.g. for `FNNLayer`, `input_size` and `output_size` could be used to construct a hash

Nothing of this needs to happen in this PR, this is just a reminder

Gerhardsa0:
I just wanna start a discussion here, so mby we should start an extra discssion thread.

Should there be an abstract Layer class, which contains all signatures of the Layer class, so the User gets better Feedback on the function calls.

We cant let the FNNLayer stay everwhere, because a neural network does not only exist out of fnn layers.

Aswell I question myself, do we want the user to change the activation function by himself?, or do we keep only internal handle.
And why do we use the softmax function for outputlayers greater 2 and sigmoid else?

sibre28:
> I just wanna start a discussion here, so mby we should start an extra discssion thread.
> 
> Should there be an abstract Layer class, which contains all signatures of the Layer class, so the User gets better Feedback on the function calls. We cant let the FNNLayer stay everwhere, because a neural network does not only exist out of fnn layers.
> 
> Aswell I question myself, do we want the user to change the activation function by himself?, or do we keep only internal handle. And why do we use the softmax function for outputlayers greater 2 and sigmoid else?

Yes, i have already added an abstract layer class. For my use-cases setting the activation functions is rather simple and can be done automatically, altough I am not against implementing the option to let the user customize them as well, it might be worth it to provide the ability to do some experimenting. Softmax and Sigmoid are only set in the Classification use-case, since we need multiple probabilitys (1 for each class) when the outputlayersize is greater than 2 and only one otherwise.

src/safeds/ml/nn/_fnn_layer.py
```diff
@@ -17,11 +19,30 @@ def __init__(self, input_size: int, output_size: int, activation_function: str):
...
                 raise ValueError("Unknown Activation Function: " + activation_function)
 
-    def forward(self, x: float) -> float:
+    def forward(self, x: Tensor) -> Tensor:
         return self._fn(self._layer(x))
 
 
-class FNNLayer:
+class Layer(ABC):
+    @abstractmethod
```
Gerhardsa0, line 27: This should be in an extra file, you should be able to use the docs from fnn_layer aswell.

tests/safeds/ml/nn/test_layer.py
lars-reimann, line 1: What's this file about?

src/safeds/ml/nn/_forward_layer.py
```diff
@@ -49,6 +50,18 @@ def __init__(self, output_size: int, input_size: int | None = None):
...
         return _InternalLayer(self._input_size, self._output_size, activation_function)
 
+    @property
+    def input_size(self) -> int:
+        """
+        Get the input_size of this layer.
+
+        Returns
+        -------
+        result :
```
Marsmaennchen221, line 60:
```suggestion
        result:
```

src/safeds/ml/nn/_model.py
```diff
@@ -57,33 +71,39 @@ def fit(
             raise OutOfBoundsError(actual=epoch_size, name="epoch_size", lower_bound=ClosedBound(1))
         if batch_size < 1:
             raise OutOfBoundsError(actual=batch_size, name="batch_size", lower_bound=ClosedBound(1))
+        if train_data.features.number_of_columns is not self._input_size:
+            raise InputSizeError(train_data.features.number_of_columns, self._input_size)
+        self._feature_names = train_data.features.column_names
+        self._batch_size = batch_size
```
Marsmaennchen221, line 77: This should only be set on the copied model so that we don't change the current model in-place

src/safeds/ml/nn/_model.py
```diff
@@ -57,33 +71,39 @@ def fit(
...
+        optimizer = torch.optim.SGD(copied_model._model.parameters(), lr=learning_rate)
+        for _ in range(epoch_size):
+            for x, y in iter(dataloader):
                 optimizer.zero_grad()
 
                 pred = copied_model._model(x)
 
                 loss = loss_fn(pred, y)
-                loss_sum += loss.item()
+                self._loss_sum += loss.item()
```
Marsmaennchen221, line 92: I think it would make more sense to reset the loss after each epoch. If you calculate the loss over the whole period of training, you cannot really compare two epochs on their loss.

src/safeds/ml/nn/_model.py
```diff
@@ -57,33 +71,39 @@ def fit(
...
                 loss.backward()
                 optimizer.step()
-                number_of_batches_done += 1
+                self._total_number_of_batches_done += 1
```
Marsmaennchen221, line 95: should not be in place

src/safeds/ml/nn/_model.py
```diff
@@ -181,51 +212,41 @@ def fit(
             raise OutOfBoundsError(actual=epoch_size, name="epoch_size", lower_bound=ClosedBound(1))
         if batch_size < 1:
             raise OutOfBoundsError(actual=batch_size, name="batch_size", lower_bound=ClosedBound(1))
+        if train_data.features.number_of_columns is not self._input_size:
+            raise InputSizeError(train_data.features.number_of_columns, self._input_size)
+        self._feature_names = train_data.features.column_names
+        self._batch_size = batch_size
```
Marsmaennchen221, line 218: should not be in-place

src/safeds/ml/nn/_model.py
```diff
@@ -181,51 +212,41 @@ def fit(
...
-                    loss = loss_fn(pred, y)
-                loss_sum += loss.item()
+
+                loss = loss_fn(pred, y)
+                self._loss_sum += loss.item()
                 loss.backward()
                 optimizer.step()
-                number_of_batches_done += 1
+
+                self._total_number_of_batches_done += 1
```
Marsmaennchen221, line 239: should not be inplace

Marsmaennchen221 (approved):

lars-reimann:
@Safe-DS/library For future PRs, please separate renamings of public API elements from internal changes. One is a minor version bump (later even a major version bump once we have a 1.x.x release), the other is a patch version bump.

PR #617, opened 2024-04-10
Imagelist hashing change

WinPlay02:
### Summary of Changes

- adapts the image list to the new common hashing infrastructure
- also resolves conflicts from merging in the master branch

This PR targets the image list branch


src/safeds/ml/classical/classification/_classifier.py
```diff
@@ -27,7 +27,7 @@ def __hash__(self) -> int:
         hash : int
```
Marsmaennchen221, line 27:
```suggestion
        hash:
```

PR #624, opened 2024-04-17
perf: lazily import our modules and external libraries

lars-reimann:
### Summary of Changes

We depend on several large libraries that take a while to load. Previously, importing `Table` would already import almost all of them, leading to horrendous startup times:

```
(Measure-Command { python -c "from safeds.data.tabular.containers import Table" }).TotalSeconds
```

➡️ 3.5068337 (seconds to import `Table` on main)

Now, we lazily import our own modules in the `__init__.py` files, and we lazily import external libraries. The latter part is quite ugly, since each function must now contain their external imports at the start. There is no better solution, however, and the improvements are huge:

```
(Measure-Command { python -c "from safeds.data.tabular.containers import Table" }).TotalSeconds
```

➡️ 0.1683219 (seconds to import `Table` in this branch)

We still have to pay the cost for an import once we first import a module, but at least this no longer has to happen fully upfront.


lars-reimann:
@Safe-DS/library For future additions, please move all imports of external libraries into the functions that need them, so they get evaluated lazily. It's hideous, but **much** faster, as explained in the opening post.

PR #625, opened 2024-04-17
feat: Add `InputConversion` & `OutputConversion` for nn interface

### LINKED ISSUE #621: feat: add Input Layer for nn Interface

We are right now working on translating the safe-ds data into pytorch tensors. The translation is defined in the dataclasses them self, but the definiton how, should be defined on the input layer interface.

### Desired solution

Create Input Layer Interface, with Type parameter, so we have static type analysis.

Gerhardsa0:
Closes #621 

### Summary of Changes

added input and output conversion Interface
also added this functionality for Table


src/safeds/ml/nn/_input_conversion.py
```diff
@@ -0,0 +1,29 @@
+from abc import ABC, abstractmethod
```
lars-reimann, line 1: Add `from __future__ import annotations`

src/safeds/ml/nn/_input_conversion.py
```diff
@@ -0,0 +1,29 @@
...
+from typing import Generic, TypeVar
+
+from torch.utils.data import DataLoader
```
lars-reimann, line 4: Move into a `TYPE_CHECKING` block.

src/safeds/ml/nn/_input_conversion_table.py
```diff
@@ -0,0 +1,27 @@
+from torch.utils.data import DataLoader
```
lars-reimann, line 1: Future/annotations + TYPE_CHECKING.

src/safeds/ml/nn/_output_conversion.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar
+
+from torch import Tensor
```
lars-reimann, line 4: Annotations, TYPE_CHECKING.

src/safeds/ml/nn/_output_conversion_table.py
```diff
@@ -0,0 +1,16 @@
+from torch import Tensor
```
lars-reimann, line 1: Annotations + TYPE_CHECKING.

src/safeds/ml/nn/_forward_layer.py
```diff
@@ -1,7 +1,7 @@
 from torch import Tensor, nn
```
lars-reimann, line 1: Should not be imported globally. This single line of code takes 1.2s to run.

src/safeds/ml/nn/_input_conversion.py
Marsmaennchen221: All of these or just the `DataLoader`?

lars-reimann: Just the `DataLoader`. It's fine to load globally

1. our stuff (`safeds...`),
2. anything built into Python.

If you add the `from __future__ import annotations`, `ruff` should also suggest moving this import into a `TYPE_CHECKING` block.

lars-reimann (changes requested):

sibre28:
We probably should add explicit test cases for the methods of InputConversionTable and OutputConversionTable. Currently these are only being covered by the model tests.

Marsmaennchen221 (approved):

Gerhardsa0:
@lars-reimann your review is needed

Issue #581, opened 2024-03-25
Remove additional softmax function if ClassificationNeuralNetwork uses CategoricalCrossEntropy


Marsmaennchen221:
### Describe the bug

Currently, the last layer of a ClassificationNeuralNetwork module uses softmax as an activation function, but this is not required and can lead to problems in the learning because pytorch CategoricalCrossEntropy already includes a softmax function, as described in the [documentation](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html):

> Note that this case is equivalent to applying [LogSoftmax](https://pytorch.org/docs/stable/generated/torch.nn.LogSoftmax.html#torch.nn.LogSoftmax) on an input, followed by [NLLLoss](https://pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html#torch.nn.NLLLoss).

### To Reproduce

/

### Expected behavior

Don't use an extra softmax function

Issue #580, opened 2024-03-22
Add CNN Layers and model support


Marsmaennchen221:
### Is your feature request related to a problem?

Add CNN Layers and model support

### Desired solution

The layers should be consistent to the already implemented fnn layer & models

PR #645, opened 2024-04-23
feat: Add `ImageDataset` and Layer for ConvolutionalNeuralNetworks

### LINKED ISSUE #579: Add a new ImageDataset Class

This ImageDataset should handle an input as an ImageList and an output as either an ImageList a column that will be one hot encoded internally or a numerical table with values between 0 and 1. If the output is a table it should have the same amount of columns as the nn module has output neurons and all columns should have values between 0 and 1. If the output is a column the amount of one hot encoded values should not exceed the amount of output neurons of the nn module.

### Desired solution

Add a new ImageDataset Class
This class should also deliver batching methods for NN models

_No response_

### Screenshots (optional)

_No response_

_No response_

Marsmaennchen221:
Closes #579, #580, #581 

### Summary of Changes

feat: added `Convolutional2DLayer`, `ConvolutionalTranspose2DLayer`, `FlattenLayer`, `MaxPooling2DLayer` and `AvgPooling2DLayer`
feat: added `InputConversionImage`, `OutputConversionImageToColumn`, `OutputConversionImageToTable` and `OutputConversionImageToImage`
feat: added generic `ImageDataset`
feat: added class `ImageSize` and methods `ImageList.sizes` and `Image.size` to get the sizes of the respective images
feat: added ability to iterate over `SingleSizeImageList`
feat: added param to return filenames in `ImageList.from_files`
feat: added option `None` for no activation function in `ForwardLayer`
feat: added `Image.__array__` to convert a `Image` to a `numpy.ndarray`
feat: added equals check to `OneHotEncoder`
fix: fixed bug https://github.com/Safe-DS/Library/issues/581 in removing the Softmax function from the last layer in `NeuralNetworkClassifier`
refactor: move `image.utils` to `image._utils`
refactor: extracted test devices from `test_image` to `helpers.devices`


src/safeds/data/image/containers/_image_dataset.py
lars-reimann, line 1: We could also put this into `src/safeds/data/labeled/_image_dataset.py`. The plan would be to also move `TaggedTable` and whatever equivalent we have for `TimeSeries` there and later add other dataset variants for, say `Image` to `Table`.

This package would then contain all the classes that can be used as inputs to supervised ML models.

Marsmaennchen221: I moved the `ImageDataset` to `src/safeds/data/labeled/containers/_image_dataset.py` in case we also need something like typing for the labeled data.

src/safeds/ml/nn/_layer.py
```diff
@@ -1,31 +1,45 @@
...
 
     @abstractmethod
-    def _get_internal_layer(self, activation_function: str) -> nn.Module:
+    def _get_internal_layer(self, **kwargs: Any) -> nn.Module:
         pass  # pragma: no cover
 
     @property
     @abstractmethod
-    def input_size(self) -> int:
+    def input_size(self) -> int | ImageSize:
```
Gerhardsa0, line 23: Why should these be ```ImageSize``` in Parent class?

src/safeds/ml/nn/_layer.py
```diff
@@ -1,31 +1,45 @@
...
         pass  # pragma: no cover
 
     @property
     @abstractmethod
-    def output_size(self) -> int:
+    def output_size(self) -> int | ImageSize:
+        pass  # pragma: no cover
```
Gerhardsa0, line 29: Why should the returntype be ```ImageSize``` in Parent class?

src/safeds/ml/nn/_input_conversion_image.py
```diff
@@ -0,0 +1,144 @@
...
+    from safeds.data.tabular.transformation import OneHotEncoder
+
+from safeds.ml.nn import InputConversion
+
+
+class InputConversionImage(InputConversion[ImageDataset, ImageList]):
+    """The input conversion for a neural network, defines the input parameters for the neural network."""
+
+    def __init__(self, image_size: ImageSize) -> None:
+        """
```
Gerhardsa0, line 23: Can we get the ``ìmage_size``` information from input_data in the Imageset?
in my PR I try to get rid of most arguments in the init function, so the user has a more easier handle.

src/safeds/ml/nn/_forward_layer.py
```diff
@@ -24,11 +26,13 @@ def __init__(self, input_size: int, output_size: int, activation_function: str):
                     self._fn = nn.ReLU()
                 case "softmax":
                     self._fn = nn.Softmax()
+                case "none":
```
Gerhardsa0, line 29: Should this be allowed? What is a use case for not having an activation function?

Marsmaennchen221: Yes, a multi-class classification uses the `CrossEntropyLoss`. This loss does already apply a `Softmax` function. Therefore, the last layer of a multi-class Classification should not have an activation function. See #581

src/safeds/ml/nn/_input_conversion_image.py
Marsmaennchen221: I need the `image_size` for the creation of the 2d layers. Currently, the pytorch model is created in the init of the model class. Therefore, the `image_size` needs to be set before the model is initialized.

Gerhardsa0 (approved):

PR #688, opened 2024-05-02
feat: improve `Column.plot_histogram`

SamanHushi:
Refactor plot_histogram() in _column

### Summary of Changes
- made plot_histogram() in _column use the funktion from _table
- updated images for Tests


SamanHushi:
I have moved my implementation to a new branchs insted of having it on my main for cleaner pullrequests in the future.
That this got closed and removed whas not intended. Should i open it with the same code again from the new branch or do nothing?

PR #709, opened 2024-05-04
ci: test on MacOS

lars-reimann:
### Summary of Changes

* Run tests on MacOS.


lars-reimann:
@Marsmaennchen221 Should we skip the failing tests on MacOS or is there a way to guarantee consistent behavior across platforms?

Marsmaennchen221:
@lars-reimann do you have a way to create snapshots on macos? Then we could generate snapshots for each os

tests/helpers/_operating_systems.py
```diff
@@ -0,0 +1,13 @@
+import platform
+import pytest
+
+os_windows = "Windows"
+os_linux = "Linux"
+os_mac = "Darwin"
+
+
+def skip_os_dependent(os: str | list[str]) -> None:
```
lars-reimann, line 10: I'd find the negation more intuitive: `_skip_if_os`

lars-reimann:
> @lars-reimann do you have a way to create snapshots on macos? Then we could generate snapshots for each os

Only in CI, but that's too cumbersome.

PR #726, opened 2024-05-05
feat: `Column.histogram()` using Table histogram funktion for consistent results

SamanHushi:
### Summary of Changes
- `Column.histogram()` now uses the same function as `Table.histograms()` and fixed that way the same problems as `Table.histograms()` had bevor the change
- update Test snapshots


lars-reimann (approved):
Looks great, thanks.

PR #741, opened 2024-05-08
feat: add continuous prediction to output conversion

### LINKED ISSUE #740: feat: add continues prediction to OutPutconvertion

Right now the outputsize of a TimeSeries in a nn Model is always 1, because we predict the value in forecast horizon steps. 

### Desired solution

I want to add the ability that the user forecasts until the forecast horizon, so the outputsize will be the forecast horizon, as all steps until the forecast horizon, will be also forecasted.

_No response_

### Screenshots (optional)

_No response_

_No response_

Gerhardsa0:
Closes #740 

### Summary of Changes

I added a continues prediction for time series.

As example w.o. continues:
we want to predict values with forecast horizon 3, we only predict one value with 3 steps into the future.
With continues, all values until the forecast horizon (forecast horizon is included) will be predicted. This brought one problem up, as we don't store list values in our Columns. So I decided to make a new column for every step in the forecast horizon. 
so with 3 there will be 3 prediction columns: prediction 0, prediction 1, prediction 2

Any Ideas on this behavior? @lars-reimann @Marsmaennchen221 @sibre28

### DIFF
--- a/src/safeds/data/labeled/containers/_time_series_dataset.py
+++ b/src/safeds/data/labeled/containers/_time_series_dataset.py
@@ -178,10 +178,9 @@ class TimeSeriesDataset:
         """
         return self._table
 
-    def _into_dataloader_with_window(self, window_size: int,
-                                     forecast_horizon: int,
-                                     batch_size: int,
-                                     continues: bool = False) -> DataLoader:
+    def _into_dataloader_with_window(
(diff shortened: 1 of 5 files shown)


lars-reimann:
The upcoming polars implementation allows storing lists inside of columns. See the experimental tabular containers.

PR #760, opened 2024-05-13
fix: test_slice_rows and fix_test_split_rows

Gerhardsa0:
### Summary of Changes

fixing the split row tests

right now, shuffle is default true, is this wanted?
Also, a empty Table created from a slice rows operation has the same schema, as the table before, so not an empty schema. Is this behavior wanted?
@lars-reimann


lars-reimann:
Thanks, I've included this in https://github.com/Safe-DS/Library/pull/762. I'll put you as a co-author there. I'll fix the remaining (30) failing tests while I rework the table transformers.

PR #767, opened 2024-05-15
feat: added from_str_to_temporal and continues prediction

### LINKED ISSUE #740: feat: add continues prediction to OutPutconvertion

### LINKED ISSUE #765: feat: add temporal operations

adding the ```time_from_str``` to library 

### Desired solution

add the functionality to library

_No response_

### Screenshots (optional)

_No response_

_No response_

### LINKED ISSUE #773: feat: change scatter and line plot, to take lists as Y argument

To be able to compare columns, there should be a functionality which allows that. As we refactored the plotting, this functionality will be implemented in the scatter and line plot, as they take now lists as their Y argument. 
### Desired solution

change the line and scatter plot function

### LINKED ISSUE #806: Remove `time_column` from `TimeSeriesDataset`

The `time_column` is never used for training. Moreover, time series analysis also works without an explicit time column (e.g. one row per day) or with multiple time columns (e.g. hour/minute/second).

### Desired solution

Remove the `time_column` from `TimeSeriesDataset`.

_No response_

### Screenshots (optional)

_No response_

_No response_

Gerhardsa0:
now we can store dates as dates and the plots can make use of the date format, see test_lineplot.

Closes #806
Closes #765 
Closes #740 
Closes #773 

### Summary of Changes

removed time name from time series class
added time series tutorial
added multiple columns for line plot and scatter plot
added temporal data interface and temporal operations, also added legend to these plots
added continuous prediction
some refactoring and fixes


lars-reimann:
> in polars they also work on series as I see. If yes, in which section should I add it.

For now, no. A column already has a method `transform`, which allows transforming the column based on an expression. Slightly more verbose, but we don't need to duplicate the API.

Gerhardsa0:
pls do not merge right now, I also work on a tutorial on this branch

lars-reimann:
I've reverted this to a draft for the time being.

Gerhardsa0:
@lars-reimann right now, the line plot creates a confidence interval for the y column.  Is there a use case for that, when there are multiple columns plotted, because it can make the plot messy. What would be preferred usage?

I have a demo implementation in table_plotter compare_plot.

lars-reimann:
The grouping is needed if there are multiple y values for the same x value. Otherwise you get a lot of crisscrossing lines.

However, we could add an option to disable the confidence intervals.

src/safeds/data/tabular/typing/_data_type.py
```diff
@@ -55,7 +55,7 @@ def is_numeric(self) -> bool:
     @abstractmethod
     def is_temporal(self) -> bool:
         """
-        Whether the column type is temporal.
+        Whether the column type is operator.
```
lars-reimann, line 58: Should not be changed.

src/safeds/data/tabular/containers/_column.py
```diff
@@ -137,7 +138,7 @@ def is_numeric(self) -> bool:
 
     @property
     def is_temporal(self) -> bool:
-        """Whether the column is temporal."""
+        """Whether the column is operator."""
```
lars-reimann, line 141: Should not be changed.

src/safeds/data/tabular/containers/_column.py
```diff
@@ -155,6 +156,11 @@ def plot(self) -> ColumnPlotter:
         """The plotter for the column."""
         return ColumnPlotter(self)
 
+    @property
+    def temporal(self) -> Temporal:
+        """Temporal operation for a column."""
+        return Temporal(self)
+
```
lars-reimann, line 163: Should only be available on cell. Columns can be transformed via the `transform` method.

Put the property next to https://github.com/Safe-DS/Library/blob/69a780cab1a5dcab33c8c7a36f03bd2eb3367683/src/safeds/data/tabular/containers/_cell.py#L145.

Maybe call it `dt`, like polars does. While we want to avoid abbreviations in general, these expressions can otherwise get very verbose.

src/safeds/data/tabular/operator/_temporal.py
```diff
@@ -0,0 +1,98 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from safeds.data.tabular.containers import Column
+
+
+class Temporal:
```
lars-reimann, line 9: Call this `TemporalCell`, the file `_temporal_cell` and move it next to `_string_cell`: https://github.com/Safe-DS/Library/blob/69a780cab1a5dcab33c8c7a36f03bd2eb3367683/src/safeds/data/tabular/containers/_string_cell.py#L12

Should operate on cells, not columns, like `StringCell`.

src/safeds/data/tabular/operator/_temporal.py
```diff
@@ -0,0 +1,98 @@
...
+        | 2024-01-01 |
+        +------------+
+        """
+        from polars import Date
+
+        from safeds.data.tabular.containers import Column
+
+        temp = self._column._series.str.strptime(Date, format_string)
+
+        return Column._from_polars_series(temp)
```
lars-reimann: That conversion is already available under `Cell.str.to_date` / `Cell.str.to_datetime`. We're generally using the `to` direction where possible.

lars-reimann (changes requested):
First batch of comments. Looks great overall. I'd appreciate more focussed PRs, though (1. plotting, 2. tutorials, 3. temporal operations).

src/safeds/data/tabular/operator/_temporal.py
lars-reimann: Feel free to add the option to specify the format string. However, that will need

* documentation,
* validation.

Gerhardsa0:
@lars-reimann the test_should_raise_value_error_when_input_date_is_invalid in the test_datetime_to_string and test_date_to_string do not work on the poetry with code cov, but I dont know why, because they work on my machine and I fixed those erros.
Any Idea why this can be?

sibre28 (dismissed):

sibre28 (approved):
[LGTM](https://lgtm.party/p/104743R/)

PR #777, opened 2024-05-17
fix: 'Table.summarize_statistics()' fixed

SamanHushi:
### Summary of Changes

cast the pl.series to str when its filled with bools when calculating variance.
This is needed because it can't get unique_counts from bools

### DIFF
--- a/src/safeds/data/tabular/containers/_column.py
+++ b/src/safeds/data/tabular/containers/_column.py
@@ -1057,7 +1057,6 @@ class Column(Sequence[T_co]):
             return 1.0  # All non-null values are the same (since there is are none)
         if type(non_missing.dtype)==pl.datatypes.Boolean: 
             non_missing = non_missing.cast(str)
-            print(non_missing)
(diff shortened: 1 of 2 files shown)


src/safeds/data/tabular/containers/_column.py
```diff
@@ -1051,10 +1051,12 @@ def stability(self) -> float:
...
         0.5
         """
+        import polars as pl
         non_missing = self._series.drop_nulls()
         if non_missing.len() == 0:
             return 1.0  # All non-null values are the same (since there is are none)
-
+        if type(non_missing.dtype)==pl.datatypes.Boolean: 
+            non_missing = non_missing.cast(str)
         mode_count = non_missing.unique_counts().max()
```
lars-reimann, line 1060: `value_counts().get_column("count").max()`

Seems to work for boolean as well, which saves us the conversion.

lars-reimann:
Thanks I've included this in https://github.com/Safe-DS/Library/pull/778 and specified you as a co-author.

PR #790, opened 2024-05-19
feat: added method to load pretrained models from huggingface

Marsmaennchen221:
### Summary of Changes

feat: added `NeuralNetworkClassifier.load_pretrained_model` and `NeuralNetworkRegressor.load_pretrained_model` to load pretrained models from huggingface. Currently supports only Image models.
feat: added `ModelImageSize`, `ConstantImageSize` and `VariableImageSize`
feat: added support for `NeuralNetworkRegressor` with images of variable size. If you use a `VariableImageSize` any image which height and/or width are a multiple of the `VariableImageSize` are being supported by the model
feat: added `NeuralNetworkClassifier.input_size` and `NeuralNetworkRegressor.input_size`
feat: changed `Column.get_distinct_values` to keep order of values in column


Marsmaennchen221:
@lars-reimann do you have an idea how to test the pretrained models? Currently, those methods are not being tested. Maybe the same approach as in Safe-DS/Datasets#164 could work, but I would appreciate it if we could do this at a later date.

lars-reimann:
> do you have an idea how to test the pretrained models?

Not right now, no.

> Currently, those methods are not being tested. Maybe the same approach as in https://github.com/Safe-DS/Datasets/pull/164 could work

We've already used all our space for caches in this project (10GB), so that won't help us.

> but I would appreciate it if we could do this at a later date.

Sure, no problem.

Gerhardsa0:
> @lars-reimann do you have an idea how to test the pretrained models? Currently, those methods are not being tested. Maybe the same approach as in [Safe-DS/Datasets#164](https://github.com/Safe-DS/Datasets/pull/164) could work, but I would appreciate it if we could do this at a later date.

For testing these models, maybe it is enough for now that we check that the structure is correct.

As example we could load an model into safe-ds and then check if its loaded correctly:
 the layers, the size of the layers, layer type etc

lars-reimann (approved):
Looks great, thanks.

PR #811, opened 2024-05-31
710 easily create a baseline model

### LINKED ISSUE #710: Easily create a baseline model

It would be helpful to easily create a baseline model to have a point of reference that must be beaten by a specifically built model.

### Desired solution

We should offer a means that trains various suitable ML models with default settings on the given data and returns the best. Maybe it should also apply basic transformations to the data (handle categorical data, handle missing data, standardize data, remove IDs).

We could create classes `BaselineRegressor` and `BaselineClassifier` that offer the usual `Regressor`/`Classifier` interface. Additionally, they should have a function to get the best model.

Tasks:

- [ ] Baseline Data Cleaning #840
- [ ] Baseline Model Training #841

_No response_

### Screenshots (optional)

_No response_

_No response_

sibre28:
Closes #710 

### Summary of Changes

Added BaselineClassifier, BaselineRegressor and test cases

### DIFF
--- a/src/safeds/ml/classical/classification/_baseline_classifier.py
+++ b/src/safeds/ml/classical/classification/_baseline_classifier.py
@@ -48,7 +48,7 @@ class BaselineClassifier:
             for model in self._list_of_model_types:
                 futures.append(executor.submit(_fit_single_model, model, train_data))
             [done, _] = wait(futures, return_when=ALL_COMPLETED)
-            for future in futures:
+            for future in done:
                 copied_model._fitted_models.append(future.result())
         executor.shutdown()
 
@@ -80,7 +80,8 @@ class BaselineClassifier:
             futures = []
             for model in self._fitted_models:
                 futures.append(executor.submit(_predict_single_model, model, test_data))
-            for future in futures:
+            [done, _] = wait(futures, return_when=ALL_COMPLETED)
+            for future in done:
                 results.append(future.result())
         executor.shutdown()
 
--- a/src/safeds/ml/classical/classification/_baseline_classifier.py
+++ b/src/safeds/ml/classical/classification/_baseline_classifier.py
@@ -1,0 +1,87 @@
+import copy
+from typing import Self
+
+from safeds.data.labeled.containers import TabularDataset
+from safeds.exceptions import ModelNotFittedError
+from safeds.ml.classical.classification import Classifier
+from safeds.ml.classical.classification import RandomForestClassifier, AdaBoostClassifier, \
+            DecisionTreeClassifier, GradientBoostingClassifier, KNearestNeighborsClassifier, SupportVectorClassifier
+
+
+def _fit_single_model(model: Classifier, train_data: TabularDataset) -> Classifier:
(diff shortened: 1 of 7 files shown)


src/safeds/ml/classical/classification/_baseline_classifier.py
```diff
@@ -0,0 +1,185 @@
...
+
+def _fit_single_model(model: Classifier, train_data: TabularDataset) -> Classifier:
+    return model.fit(train_data)  # pragma: no cover
+
+
+def _predict_single_model(model: Classifier, test_data: TabularDataset) -> TabularDataset:
+    return model.predict(test_data)  # pragma: no cover
+
+
+class BaselineClassifier:
```
Gerhardsa0, line 31: is there still a need to differ between a classifier and a regression?

src/safeds/ml/classical/classification/_baseline_classifier.py
```diff
@@ -0,0 +1,185 @@
...
+        from safeds.ml.metrics import ClassificationMetrics
+
+        if not self._is_fitted:
+            raise ModelNotFittedError
+
+        # Validate data
+        if not self._feature_names == test_data.features.column_names:
+            raise FeatureDataMismatchError
+        if not self._target_name == test_data.target.name:
+            raise DatasetMissesTargetError(self._target_name)
```
Gerhardsa0, line 143: maybe rename this error in DatasetTargetDoesNotMatch or something in that direction
The name suggest, that a target column is missing, but they are not matching

src/safeds/ml/classical/regression/_baseline_regressor.py
```diff
@@ -0,0 +1,199 @@
...
+)
+
+
+def _fit_single_model(model: Regressor, train_data: TabularDataset) -> Regressor:
+    return model.fit(train_data)  # pragma: no cover
+
+
+def _predict_single_model(model: Regressor, test_data: TabularDataset) -> TabularDataset:
+    return model.predict(test_data)  # pragma: no cover
+
```
Gerhardsa0, line 33: why are they not tested?

src/safeds/ml/classical/classification/_baseline_classifier.py
sibre28: The only differences between them are the list of Models and the list of metrics. We do need the user to explicitly tell us which one to use, as we currently do not have a way to differentiate the two only by the given data. We could change the current implementation to just one Baseline Class and add a parameter to define the use case but given our other model implementations I would say it makes sense to keep two classes for consistency.

src/safeds/ml/classical/regression/_baseline_regressor.py
sibre28: They are being implicitly tested, as they are called a bunch of times in the other Baseline tests. Codecov doesn't seem to recognize this, probably because we are starting new processes and calling them there.

Gerhardsa0 (approved):

Gerhardsa0 (approved):

PR #838, opened 2024-06-12
feat: prediction no longer takes a time series dataset only table

### LINKED ISSUE #837: feat: change that nn takes no time series for predict

right now a nn is trained with a time series and also predicted with a time series. But these timeseries need to be the same so we can just take the information from the trained model and keep it for prediction.

### Desired solution

the prediction should just take a table.

Gerhardsa0:
Closes #837

### Summary of Changes

Prediction no longer takes time series dataset.

Should the model really output the same datatype as the fit data (look abstract class InputConversion), for my Use Case I think it is more intuitive to output a Table instead of a TimeSeriesDataset.


Gerhardsa0:
@lars-reimann allowing Dataset as type for predict would take a deeper impact to the interface, as the table interface and Image the image interface also need to allow dataset as type. Is this wanted?

lars-reimann:
Let's defer that change until later and only allow tables for now.

src/safeds/ml/nn/converters/_input_converter_time_series.py
```diff
@@ -24,9 +24,9 @@ def __init__(
         self._forecast_horizon = 0
         self._first = True
         self._target_name: str = ""
-        self._time_name: str = ""
         self._feature_names: list[str] = []
         self._continuous: bool = False
+        self._extra_names: list[str] = []
```
Marsmaennchen221, line 29: Add the `_extra names` in the `__eq__`, `__hash__` and `__sizeof__` methods

Marsmaennchen221 (changes requested):

Marsmaennchen221 (approved):

PR #843, opened 2024-06-19
feat: hyperparameter optimization for classical models

### LINKED ISSUE #264: Hyperparameter optimization

Finding appropriate values for hyperparameters by hand is tedious. There should be automation to try different combinations of values.

### Desired solution

1. For all hyperparameters of models of type `T` it should also be possible to pass a `Choice[T]` (see #325). Example:
  ```py
  # Before
  class KNearestNeighbors(Classifier):
    def __init__(self, number_of_neighbors: int) -> None:
      ...
  
  # After
  class KNearestNeighbors(Classifier):
    def __init__(self, number_of_neighbors: int | Choice[int]) -> None:
      ...

  # Usage
  KNearestNeighbors(number_of_neighbors = Choice(1, 10, 100))
  ```
2. Adjust the getters (#260) accordingly.
3. When a user tries to call `fit` on a model that contains `Choice` at any level (can be nested), raise an exception. Also point to the correct method (see 4.).
4. Add new method `fit_by_exhaustive_search` to `Classifier` and subclasses with parameter:
    * `optimization_metric`: The metric to use to find the best model. It should have type `ClassifierMetric`, which is an enum with one value for each classifier metric we have available:
    ```py
    class ClassifierMetric(Enum):
        ACCURACY = "accuracy"
        PRECISION = "precision
        RECALL = "recall"
        F1_SCORE = "f1_score"
    ```
    The parameter should be required.
5. Add new method `fit_by_exhaustive_search` to `Regressor` and subclasses with parameter:
    * `optimization_metric`: The metric to use to find the best model. It should have type `RegressorMetric`, which is an enum with one value for each regressor metric we have available:
    ```py
    class RegressorMetric(Enum):
        MEAN_SQUARED_ERROR = "mean_squared_error"
        MEAN_ABSOLUTE_ERROR = "mean_absolute_error"
    ```
    The parameter should be required.
6. Both of those methods should then collect the `Choice`s inside of the model and its children, and for each possible setting create a model without choices, fit this, and compute the listed metric on it. It should then keep track of the best (fitted) model according to the metric and return it at the end. [`GridSearchCV`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html) of `scikit-learn` can be useful for this.

sibre28:
Closes #264 

Adjusted classical ML-Models to support taking a Choice Parameter

New Features for Classifiers and Regressors:
- combined Linear,Lasso,Ridge and ElasticNetRegressor into ElasticNetRegressor
- changed property methods and parameter types
- added fit_by_exhaustive_search(), to fit a model with all combinations of given Choices
- added Errors for using the wrong fit method (fit with or fit_by_exhaustive_search without Choice Parameter)
- added Enums ClassifierMetric and RegressorMetric, which are passed to fit_by_exhaustive_search to determine the optimization metric
- added cross validation in fit_by_exhaustive_search
- added multiprocessing in fit_by_exhaustive_search

- added tests for all methods and classes


sibre28:
Related Bug Fixes:

https://github.com/PrefectHQ/prefect/issues/10794

Gerhardsa0 (approved):

PR #846, opened 2024-06-20
feat: add `ImageDataset.split`

### LINKED ISSUE #831: split ImageList to generate train and test sets

we wanted to generate a test set but couldn't do it so quickly and also for some reason when we created our own image but its had some confusion issues till it worked

### Desired solution

it should have a spit function like 
```python
def split_imageList(self, split_ratio: float)-> tuple[train: ImageList, test: ImageList]:
# it should convert the ImageList into List[Image] with ImageList.to_images() 
# and then shuffle and split the list and use ImageList.from_images(image_list) 
# to convert them both back into an ImageList and return the tuple```

_No response_

### Screenshots (optional)

_No response_

_No response_

Marsmaennchen221:
Closes #831 

### Summary of Changes

feat: add `ImageDataset.split`


lars-reimann (approved):
👍

Issue #849, opened 2024-06-21
`plot.box_plots` with matplotlib


lars-reimann:
### Is your feature request related to a problem?

`plot.box_plots` really is the last plot method that still uses seaborn. It should be rewritten using matplotlib, so we can remove the dependency on seaborn.

### Desired solution

/

PR #850, opened 2024-06-21
feat: replaced seaborn with matplotlib for correlation_heatmap

### LINKED ISSUE #800: `plot.correlation_heatmap` with matplotlib
`plot.correlation_heatmap` is the last plot method that still uses seaborn. It should be rewritten using matplotlib, so we can remove the dependency on seaborn.

Muellersen:
Closes #800 

### Summary of Changes

Removed seaborn with matplotlib for correlation_heatmap.
Matplotlib returns a slightly different plot, so we also replaced the snapshot image.

But seaborn still remains as a dependency, because box_plots is still using seaborn, see issue #849 .


lars-reimann (approved):
Looks better than before, thanks!

PR #852, opened 2024-06-21
feat: raise if `remove_colums` is called with unknown column by default

### LINKED ISSUE #807: Raise if `remove_columns` is called with unknown column by default

Currently, `Table.remove_columns` ignores unknown columns by default. However, then it can happen that users expect that certain columns are removed, but they are still there due to a typo.

### Desired solution

Add a new **optional, keyword-only** parameter `ignore_unknown_names: bool = False`.

If `ignore_unknown_names` is `False`, call `_check_columns_exist` to ensure that all given names are valid. Otherwise, keep the current behavior.

_No response_

### Screenshots (optional)

_No response_

_No response_

TobiasPluecker:
Closes #807

### Summary of Changes

- Added an optional, keyword-only parameter ignore_unknown_names: bool = False to the remove_columns method.
- This parameter controls whether an error is raised when attempting to remove non-existent columns.
- If ignore_unknown_names is set to False, the method checks for the existence of specified columns.
- If ignore_unknown_names is set to True, non-existent columns are ignored, and no error is raised.


src/safeds/data/tabular/containers/_table.py
```diff
@@ -642,15 +642,15 @@ def remove_columns(
...
         """
         Return a new table without the specified columns.
 
         **Notes:**
 
         - The original table is not modified.
-        - This method does not raise if a column does not exist. You can use it to ensure that the resulting table does
-          not contain certain columns.
 
         Parameters
```
lars-reimann, line 655: The new parameter should be documented here.

lars-reimann (changes requested):

lars-reimann:
Since `remove_columns` is also used internally, you need to adjust some additional code and set `ignore_unknown_names` to `True`.

lars-reimann (approved):
👍

PR #864, opened 2024-06-28
feat: add KNearestNeighborsImputer

### LINKED ISSUE #743: Add `KNearestNeighborsImputer`

We currently only have a basic imputer, but there are more sophisticated imputation strategies.

### Desired solution

Add a `KNearestNeighborsImputer` that uses [the KNNImputer of scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.impute.KNNImputer.html) internally.

* Superclass: `TableTransformer`
* Constructor parameters:
  * `neighbor_count: int`,
  * `column_names: str | list[str] | None = None` (keyword-only). List of columns to transform, if None all columns passed in `fit`.
  * `value_to_replace: float | str | None = None` (keyword-only)
* Attributes:
  * `self._wrapped_transformer: sk_KNNImputer | None = None`
* `fit`:
  * Call `_check_columns_exist` to ensure columns to transform exist
  * Raise value error if `row_count` is 0
  * Create a new instance of the `KNearestNeighborsImputer`, don't mutate it in place
  * Create and fit an `sk_KNNImputer` and store it in `_wrapped_transformer` of the copied transformer
* `transform`:
  * `TransformerNotFittedError` if transformer is not fitted
  * Call `_check_columns_exist` to ensure columns to transform exist

  * Transform with the `_wrapped_transformer`

_No response_

### Screenshots (optional)

_No response_

_No response_

SamanHushi:
Closes #743 

### Summary of Changes

Added KNN Imputer and Tests for it


SamanHushi:
how should we test __hash__() when its used everywhere?
The imputer discredizer also uses __hash__() but also don't cover it by a specific test

SamanHushi:
how should we cover the __hash__() ?
We don't see a test in discretiser for it where __hash__() also got used

lars-reimann:
Have a look at how `hash` was tested for `Table`: https://github.com/Safe-DS/Stdlib/blob/6fbe5379ae7676906bf8d0f21fe00926d792c60f/tests/safeds/data/tabular/containers/_table/test_hash.py

LIEeOoNn:
i also added a check if `value_to_replace ` is `None` we want to replace it with `np.nan` so the sklearn KNN_Imputer can handle the parameters correctly without throwing an error.
It will still impute values =`None` with the calculated value!

src/safeds/data/tabular/transformation/_k_nearest_neighbors_imputer.py
```diff
@@ -0,0 +1,157 @@
...
+    # ------------------------------------------------------------------------------------------------------------------
+    # Dunder methods
+    # ------------------------------------------------------------------------------------------------------------------
+
+    def __init__(
+        self,
+        neighbor_count: int,
+        *,
+        column_names: str | list[str] | None = None,
+        value_to_replace: float | str | None = nan,
```
lars-reimann, line 41:
```suggestion
        value_to_replace: float | str | None = None,
```

The user should not see that `nan` is used internally.

src/safeds/data/tabular/transformation/_k_nearest_neighbors_imputer.py
```diff
@@ -0,0 +1,157 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from numpy import nan
```
lars-reimann, line 5: For performance reasons, avoid top-level imports of third-party libraries. Due to the eager import system of Python, a simple import like this can take several seconds. To emulate lazy imports, move the import into the function that needs the `nan` (here `__init__`).

lars-reimann (changes requested):
Nice work, just two small changes left.

src/safeds/data/tabular/transformation/_k_nearest_neighbors_imputer.py
```diff
@@ -0,0 +1,157 @@
...
+    @property
+    def is_fitted(self) -> bool:
+        """Whether the transformer is fitted."""
+        return self._wrapped_transformer is not None
+
+    @property
+    def neighbor_count(self) -> int:
+        """The number of neighbors to consider when imputing missing values."""
+        return self._neighbor_count
+
```
lars-reimann, line 75: An accessor for `value_to_replace` is missing. Since the user should not get `nan`, you are better off moving the replacement of `None` with `nan` out of `__init__` and into `fit`.

src/safeds/data/tabular/transformation/_k_nearest_neighbors_imputer.py
```diff
@@ -108,6 +110,10 @@ def fit(self, table: Table) -> KNearestNeighborsImputer:
         else:
             column_names = self._column_names
             _check_columns_exist(table, column_names)
+        
+        if self._value_to_replace is None:
+            from numpy import nan
+            self._value_to_replace = nan
```
lars-reimann, line 118: You need a new local variable, rather than assigning to the field. Otherwise, this becomes visible when accessing `value_to_replace` after fitting.

lars-reimann (changes requested):

lars-reimann (approved):
Nice contribution, thanks!

PR #865, opened 2024-06-28
feat: Regularization strength for logistic (regression) classifier

### LINKED ISSUE #750: Regularization strength for logistic (regression) classifier

There should be a way to control regularization of the `LogisticClassifier`.

### Desired solution

Add an **optional, keyword-only** constructor parameter `c: float = 1.0` and pass it to the wrapped scikit-learn estimator.

I've originally also considered letting the user choose the penalty ("l1"/"l2"/"elasticnet"). However, most solvers only support "l2" anyway. We should rather stick to "l2" and later choose an appropriate solver internally based on the shape of the data for fast convergence (different issue).

### Screenshots (optional)

_No response_

_No response_

zogomii:
Closes #750


lars-reimann:
Closed since the brances were the wrong way round.

PR #879, opened 2024-06-29
perf: improved performance in various methods in `Image` and `ImageList`

Marsmaennchen221:
### Summary of Changes

1. Improved memory usage and runtime in:

- `Image`
  - `convert_to_grayscale`
  - `adjust_brightness`
  - `add_noise`
  - `adjust_contrast`
  - `adjust_color_balance`
  - `find_edges`
- `ImageList`
  - `from_images`
  - `add_image`
  - `add_images`
  - `remove_image_by_index`
  - `remove_duplicate_images`
  - `convert_to_grayscale`
  - `resize`
  - `crop`
  - `adjust_brightness`
  - `add_noise`
  - `adjust_contrast`
  - `adjust_color_balance`
  - `find_edges`

2. Changed `blur` algorithm in `Image` and `ImageList` from Gaussian blur to box blur
3. Fixed a bug in `blur` and `sharpen`, that they could not work with Tensors of size greater than 2**31

### Details to the performance upgrades:

_These details will explain the performance upgrades in the `ImageList`. All performance upgrades and changes in `Image` are made according to the changes in `ImageList`_

#### Early stopping

`convert_to_grayscale` returns `self` when it has only one channel (in this case, the `ImageList` is already in grayscale)
`remove_duplicate_images` returns `self` if the unique image tensor has the same size as the original (in this case, there are no duplicates)
`adjust_brightness` returns `ImageList` with complete black images if factor is 0
`adjust_color_balance` returns `ImageList.convert_to_grayscale` if factor is 0

#### General changes

If a float Tensor is used during the computation, it will be `float16` instead of `float32`
Improved the order of Tensor allocations, so that there will be fewer problems with tensors not being completely on the VRAM

#### Benchmark

_Only the transformation methods have benchmarks for the runtime. Their benchmark includes only changes over/under a change factor of 0.25, rounded to one decimal point, as runtime depends on multiple factors and fluctuates heavily with changes of only a few milliseconds in most cases._
_All differences are measured as a factor compared to the original results. That means a factor below 1 is worse, while a factor above 1 is a better result. For readability, all factors equal to 1 are not included._
_Due to the bug fix mentioned above for `blur` and `sharpen` the performance of these methods decreased in most cases._

##### Benchmark with RGB images of size 250×250
| method                  | result size difference | runtime difference | max memory allocation during runtime difference |
|-------------------------|------------------------|--------------------|-------------------------------------------------|
| `from_images`           | 8                      |                    | 7,95                                            |
| `remove_image_by_index` |                        |                    | 1,5                                             |
| `adjust_brightness`     |                        | 1,6                | 1,5                                             |
| `add_noise`             | 4                      | 42,5               | 2,25                                            |
| `adjust_contrast`       |                        | 1,4                | 1,69                                            |
| `adjust_color_balance`  | 4                      | 6,3                | 3                                               |
| `blur`                  |                        | 0,7                | 0,9                                             |
| `sharpen`               |                        | 0,6                | 0,91                                            |
| `find_edges`            |                        | 1,3                |                                                 |

##### Benchmark with RGBA images (RGB images with transparent layer) of size 256×256
| method                  | result size difference | runtime difference | max memory allocation during runtime difference |
|-------------------------|------------------------|--------------------|-------------------------------------------------|
| `from_images`           | 8                      |                    | 6,29                                            |
| `remove_image_by_index` |                        |                    | 1,5                                             |
| `adjust_brightness`     |                        | 1,7                | 1,36                                            |
| `add_noise`             | 4                      | 24,5               | 2,25                                            |
| `adjust_contrast`       |                        |                    | 1,53                                            |
| `adjust_color_balance`  | 4                      | 1,7                | 2,6                                             |
| `blur`                  |                        | 1,6                | 0,9                                             |
| `sharpen`               |                        |                    | 0,89                                            |

##### Benchmark with RGB images of multiple different sizes
| method                  | result size difference | runtime difference | max memory allocation during runtime difference |
|-------------------------|------------------------|--------------------|-------------------------------------------------|
| `from_images`           | 7,7                    |                    | 6,3                                             |
| `add_images`            | 8                      |                    | 2,38                                            |
| `remove_image_by_index` |                        |                    | 1,28                                            |
| `resize`                |                        | 1,3                |                                                 |
| `crop`                  |                        | 1,3                | 1,09                                            |
| `adjust_brightness`     |                        | 1,3                | 1,09                                            |
| `add_noise`             | 4                      | 44,3               | 1,47                                            |
| `adjust_contrast`       |                        | 2                  | 2,17                                            |
| `adjust_color_balance`  | 4                      | 6,7                | 1,65                                            |
| `blur`                  |                        |                    | 2,87                                            |
| `sharpen`               |                        | 0,7                | 0,9                                             |
| `find_edges`            |                        | 1,9                | 0,91                                            |


lars-reimann:
Do you have benchmarks to quantify the performance improvement?

lars-reimann (approved):
Nice performance improvements, great work.

PR #891, opened 2024-07-02
feat: activation function for forward layer

### LINKED ISSUE #889: feat: allow activation function parameter for forward layer

I need a layer without an activation function, but right now there is no way to specify the function in any column.

### Desired solution

Add a optional parameter for the forward layer

_No response_

### Screenshots (optional)

_No response_

_No response_

Gerhardsa0:
Closes #889

### Summary of Changes

added optional activation function parameter to the forward layer


sibre28:
It might be worth implementing this as an Enum so Users can see the options and it eliminates typos

lars-reimann:
> It might be worth implementing this as an Enum so Users can see the options and it eliminates typos

I don't think we've used Python's enums anywhere so far. They don't offer major benefits over literal types.

tests/safeds/ml/nn/layers/test_forward_layer.py
```diff
@@ -177,3 +177,49 @@ def test_should_assert_that_different_forward_layers_have_different_hash(
...
+
+@pytest.mark.parametrize(
+    "activation_function",
+    [
+        "a",
+        "b",
+        "c",
+        "d",
+    ],
+    ids=["sigmoid", "relu", "softmax", "none"],
```
sibre28, line 214: Reduce to one unknown string, there is no need to test this four times. Put something like "unknown" as id

src/safeds/ml/nn/layers/_forward_layer.py
```diff
@@ -21,18 +21,25 @@ class ForwardLayer(Layer):
...
+        If the given activation function does not exist
     """
 
-    def __init__(self, neuron_count: int):
+    def __init__(self, neuron_count: int, activation_function: str = "notset"):
         _check_bounds("neuron_count", neuron_count, lower_bound=_ClosedBound(1))
 
         self._input_size: int | None = None
         self._output_size = neuron_count
+        self._activation_function: str = activation_function
```
sibre28, line 42: Include self._activation_function in __hash__, __sizeof__ and __eq__

sibre28 (changes requested):

src/safeds/ml/nn/layers/_forward_layer.py
```diff
@@ -21,18 +21,25 @@ class ForwardLayer(Layer):
     neuron_count:
         The number of neurons in this layer
```
Marsmaennchen221, line 23: Remove empty line

src/safeds/ml/nn/layers/_forward_layer.py
```diff
@@ -21,18 +21,25 @@ class ForwardLayer(Layer):
...
 
+    activation_function:
+        The activation function used in the forward layer
+
     Raises
     ------
     OutOfBoundsError
         If input_size < 1
         If output_size < 1
+
```
Marsmaennchen221, line 32:
```suggestion
```

src/safeds/ml/nn/layers/_forward_layer.py
```diff
@@ -21,18 +21,25 @@ class ForwardLayer(Layer):
...
+    ValueError
+        If the given activation function does not exist
     """
 
-    def __init__(self, neuron_count: int):
+    def __init__(self, neuron_count: int, activation_function: str = "notset"):
```
Marsmaennchen221, line 37: If I understand correctly, if this variable is anything other than `"notset"` the default activation function will be overwritten. Therefore, I would call the parameter something like `overwrite_activation_function` to clarify this behavior. Similar to this, I would clarify this behavior more clear in the doc string. Also, please do not use `str` but `Literal` to specify the allowed activation functions.

Marsmaennchen221 (changes requested):

src/safeds/ml/nn/layers/_forward_layer.py
Gerhardsa0: added

src/safeds/ml/nn/layers/_forward_layer.py
Gerhardsa0: added

src/safeds/ml/nn/layers/_forward_layer.py
Gerhardsa0: added

tests/safeds/ml/nn/layers/test_forward_layer.py
Gerhardsa0: added

src/safeds/ml/nn/layers/_forward_layer.py
Gerhardsa0: added

tests/safeds/ml/nn/layers/test_forward_layer.py
```diff
@@ -177,3 +177,37 @@ def test_should_assert_that_different_forward_layers_have_different_hash(
...
+    forward_layer: ForwardLayer = ForwardLayer(1)
+    match activation_function:
+        case "sigmoid":
+            forward_layer = ForwardLayer(neuron_count=1, overwrite_activation_function="sigmoid")
+        case "relu":
+            forward_layer = ForwardLayer(neuron_count=1, overwrite_activation_function="relu")
+        case "softmax":
+            forward_layer = ForwardLayer(neuron_count=1, overwrite_activation_function="softmax")
+        case "none":
+            forward_layer = ForwardLayer(neuron_count=1, overwrite_activation_function="none")
```
sibre28, line 202: Is there a reason for the switch statement? Otherwise Simplify to
```suggestion
        forward_layer = ForwardLayer(neuron_count=1, overwrite_activation_function=activation_function)
```

Gerhardsa0: yes otherwise mypy throws an error, because a string is not a valid literal

sibre28: Try changing the type of the activation_function parameter from str to Literal["sigmoid","relu", ....]

Gerhardsa0: changed

sibre28 (dismissed):

Marsmaennchen221 (dismissed):

src/safeds/data/tabular/plotting/_table_plotter.py
```diff
@@ -572,6 +572,8 @@ def moving_average_plot(
                 ylabel=y_name,
             )
             ax.legend()
+            if self._table.get_column(x_name).is_temporal and self._table.get_column(x_name).row_count<9:
+                ax.set_xticks(x_data)  # Set x-ticks to the x data points
```
Marsmaennchen221: Could you test plotting of data with more than 8 rows?

Gerhardsa0: added

src/safeds/data/tabular/plotting/_table_plotter.py
```diff
@@ -572,7 +572,7 @@ def moving_average_plot(
                 ylabel=y_name,
             )
             ax.legend()
-            if self._table.get_column(x_name).is_temporal:
+            if self._table.get_column(x_name).is_temporal and self._table.get_column(x_name).row_count < 9:
```
Marsmaennchen221, line 575: What difference does the added part of this line do? If I try it without the row count I get the same result

Gerhardsa0: this is for rows way greater than 9, when we have 1000 rows it tries to set as many dates as possible for the dates and the plots get messie

sibre28 (approved):

src/safeds/data/tabular/plotting/_table_plotter.py
Marsmaennchen221: So the idea is to limit the labels to a maximum of 8?

Issue #898, opened 2024-07-11
Add Numpy 2.0 support


Marsmaennchen221:
Currently, we only support Numpy versions below 2.0.

torchvision will support Numpy 2.0 with version 0.19.0 (numpy/numpy#26191) wich will become available ~24th July

lars-reimann:
Done in #929.

PR #899, opened 2024-07-11
build: use numpy<2.0.0

Marsmaennchen221:
### Summary of Changes

Add clause to `pyproject.toml` to use a numpy version below 2.0 as torchvision does not yet support numpy 2.0 (See #898)


lars-reimann:
Why do we need to pin the version of transitive dependencies? Doesn't torchvision declare itself that it needs major version 1 for numpy?

Marsmaennchen221:
> Why do we need to pin the version of transitive dependencies? Doesn't torchvision declare itself that it needs major version 1 for numpy?

torchvision does not specify any restrictions on the numpy version and as of pytorch version 2.3.0 pytorch does support numpy v2.0. While torchvision claims to be 98% compatible to Numpy v2.0, a few important methods like `read_image` or `save_image` do not support v2.0. So if you install the most recent version of safe-ds you will get numpy 2.0, but a few key methods like `Image.from_file` will not work.

lars-reimann:
> torchvision does not specify any restrictions on the numpy version

😠

lars-reimann (approved):

PR #901, opened 2024-07-12
feat: add FunctionalTableTransformer

### LINKED ISSUE #858: Functional table transformer

Currently, users can transform a table either with a `TableTransformer` or with the methods of `Table`. However, only a `TableTransformer` can be used in a `SequentialTableTransformer` (see #802).

### Desired solution

* Add a new class `FunctionalTableTransformer`
* **Superclass:** `TableTransformer`
* **Constructor parameters:**
  * `transformer: Callable[[Table], Table]` The function transforms a table and returns the transformed table.
* `is_fitted`: Always `True`, immediately after constructing the model.
* `fit`: Is a no-OP. Just return `self`.
* `transform`: Call the `transformer` you got in the constructor and return the result.

_No response_

### Screenshots (optional)

_No response_

_No response_

Tarmandan:
Closes #858 

### Summary of Changes

This implements a Transformer that wraps a Callable[[Table], Table] so operations on Tables can be inserted into the operation order of a SequentialTableTransformer.
Note that this transformer inherently cannot be invertible.
Since there is no type checking at runtime, callables with wrong argument or return types will only throw an exception upon calling transform but not during init.

### DIFF
--- a/src/safeds/data/tabular/transformation/_functional_table_transformer.py
+++ b/src/safeds/data/tabular/transformation/_functional_table_transformer.py
@@ -26,13 +26,12 @@ class FunctionalTableTransformer(TableTransformer):
     # Dunder methods
     # ------------------------------------------------------------------------------------------------------------------
 
-    def __init__(self,
-                 transformer: Callable[[Table], Table],
-                 ) -> None:
+    def __init__(
+        self,
+        transformer: Callable[[Table], Table],
+    ) -> None:
         super().__init__(None)
         self._transformer = transformer
-
-        
 
     def __hash__(self) -> int:
         return _structural_hash(
--- a/src/safeds/data/tabular/transformation/_functional_table_transformer.py
+++ b/src/safeds/data/tabular/transformation/_functional_table_transformer.py
@@ -92,11 +92,7 @@ class FunctionalTableTransformer(TableTransformer):
             Raised when the wrapped callable encounters an error.
         
         """
-        try:
-            return self._transformer(table)
-        except Exception as e:
-            #TODO Evaluate if switch to non-generic exception is useful, as _func can be any callable
-            raise Exception("The underlying function encountered an error") from e  # noqa: TRY002
+        return self._transformer(table)
 
     def fit_and_transform(self, table: Table) -> tuple[FunctionalTableTransformer, Table]:
         """
--- a/src/safeds/data/tabular/transformation/_functional_table_transformer.py
+++ b/src/safeds/data/tabular/transformation/_functional_table_transformer.py
@@ -86,11 +86,16 @@ class FunctionalTableTransformer(TableTransformer):
 
         Raises
         ------
-        #TODO Implement Errors from the table methods
+        Exception:
+            Raised when the called _func encounters an error.
+        #TODO Switch to non-generic exception
         
         """
-        transformed_table = self._func.__call__(table)
-        return transformed_table
+        try:
+            transformed_table = self._func.__call__(table)
+            return transformed_table
+        except Exception as e:
+            raise Exception("The underlying function encountered an error") from e
 
     def fit_and_transform(self, table: Table) -> tuple[Self, Table]:
         """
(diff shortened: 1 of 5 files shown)


Tarmandan:
The linter found 2 problems. I fixed the first by adding type hint to the funciton valid_callable in the tests.
However the other problem is that the linter correctly notices that invalid_callable has the wrong types. I am not sure if the test is superfluous. The test for internal error of the callable raises an exception in the same try-except, just later.

src/safeds/data/tabular/transformation/_functional_table_transformer.py
```diff
@@ -0,0 +1,119 @@
...
+    funct:
+        The Callable that receives a table and returns a table that is to be wrapped.
+    """
+
+    # ------------------------------------------------------------------------------------------------------------------
+    # Dunder methods
+    # ------------------------------------------------------------------------------------------------------------------
+
+    def __init__(self,
+                 funct: Callable[[Table], Table],
```
lars-reimann:
```suggestion
                 transformer: Callable[[Table], Table],
```

See issue. Use a rename refactoring. Don't forget to update the docstring.

src/safeds/data/tabular/transformation/_functional_table_transformer.py
```diff
@@ -0,0 +1,119 @@
...
+                 ) -> None:
+        super().__init__(None)
+        self._func = funct
```
lars-reimann, line 34: Rename attribute to `self._transformer`.

src/safeds/data/tabular/transformation/_functional_table_transformer.py
```diff
@@ -0,0 +1,119 @@
...
+    from collections.abc import Callable
+
+    from safeds.data.tabular.containers import Table
+
+from ._table_transformer import TableTransformer
+
+
+class FunctionalTableTransformer(TableTransformer):
+    """
+    Learn a transformation for a set of columns in a `Table` and transform another `Table` with the same columns.
```
lars-reimann, line 17: Should be updated to better reflect the purpose of this class.

src/safeds/data/tabular/transformation/_functional_table_transformer.py
```diff
@@ -0,0 +1,118 @@
...
+        ------
+        Exception:
+            Raised when the wrapped callable encounters an error.
+
+        """
+        try:
+            return self._func(table)
+        except Exception as e:
+            # TODO Evaluate if switch to non-generic exception is useful, as _func can be any callable
+            raise Exception("The underlying function encountered an error") from e  # noqa: TRY002
```
lars-reimann, line 98: Would also be fine to remove the wrapping into another `Exception` completely and just let the original exception bubble up. This addtional exception does not add useful information.

lars-reimann (changes requested):

Tarmandan:
The test for invalid callable type has been removed.
The only problem remaining is test coverage. The __hash__ test in test_table_transformer.py should be adjusted, but FunctionalTableTransformer should only give different hash when wrapping different callable, so it will need its own tests. It might be less work to implement it in the test_functional_table_transformer.py, but would break the pattern.

lars-reimann:
For now, it should suffice to add the transformer to https://github.com/Safe-DS/Stdlib/blob/69a780cab1a5dcab33c8c7a36f03bd2eb3367683/tests/safeds/data/tabular/transformation/test_table_transformer.py#L35. We can adjust the test later so `hash` is tested more in-depth for all transformers. Maybe add an issue for that.

src/safeds/data/tabular/transformation/_functional_table_transformer.py
Tarmandan: Done

src/safeds/data/tabular/transformation/_functional_table_transformer.py
Tarmandan: Done

src/safeds/data/tabular/transformation/_functional_table_transformer.py
Tarmandan: Wrapper has been removed

src/safeds/data/tabular/transformation/_functional_table_transformer.py
Tarmandan: Text has been adjusted, please check if adequate

lars-reimann (approved):
Thanks, nice work.

